PR #129, opened 2022-11-18
Run unit tests in CI

mohsen1:


.github/workflows/ci.yml
```diff
@@ -13,4 +13,15 @@ jobs:
...
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/setup-node@v3
+          with:
+            node-version: 16
+        - uses: actions/checkout@v2
+        - name: Install modules
+          uses: bahmutov/npm-install@v1
+        - name: Run unit tests
+          run: yarn test
```
grzpab, line 27: can you ensure a time limit on the server (let's say 20s)?

mohsen1: the smallest is 1 minute. I don't think it's necessary as it's running super fast right now

gregintuitaio: do you think it makes sense to add it so we never run out of credits?

mohsen1: for open source it's all free.

PR #192, opened 2023-03-28
Feature/int 692 add ability to pass user id from web app

DmytroHryshyn:


src/components/user/userAccountStorage.ts
```diff
@@ -0,0 +1,14 @@
+import * as vscode from 'vscode';
+
+export class GlobalStateAccountStorage {
+	constructor(private readonly globalState: vscode.Memento) {}
```
grzpab, line 4: I usually write it like `__globalState`

src/components/user/userService.ts
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+export interface UserAccountStorage {
+	getUserAccount(): string | null;
+	setUserAccount(value: string | undefined): void;
+}
+
+export class UserService {
+	constructor(private readonly storage: UserAccountStorage) {}
```
grzpab, line 13: `__storage`

src/extension.ts
```diff
@@ -928,27 +935,34 @@ export async function activate(context: vscode.ExtensionContext) {
...
+					vscode.window.showTextDocument(document);
+				}
 
-				vscode.window.showTextDocument(document);
+				if (userId) {
+					try {
+						userService.linkUsersIntuitaAccount(userId);
+					} catch (e) {
+						if (e instanceof AlreadyLinkedError) {
+							// @TODO show warning asking to replace with new account
```
grzpab, line 962: I think you can use window.showNotification or similar here.

src/extension.ts
```diff
@@ -928,27 +935,34 @@ export async function activate(context: vscode.ExtensionContext) {
 			handleUri: async (uri) => {
 				const searchParams = new URLSearchParams(uri.query);
 				const base64UrlEncodedContent = searchParams.get('c');
+				const userId = searchParams.get('userId');
```
grzpab, line 938: does it need to be base64urlencoded?

src/components/user/userAccountStorage.ts
DmytroHryshyn: so, for all private vars we should use `__` prefix, right?

src/extension.ts
DmytroHryshyn: i think no

grzpab (approved):

PR #193, opened 2023-03-28
Feat/int 690 show error messages from the codemod

henoktsegayetoptal:


src/components/engineService.ts
```diff
@@ -261,10 +270,49 @@ export class EngineService {
 			shell: true,
 		});
 
-		childProcess.stderr.on('data', (data) => {
-			console.error(data.toString());
+		const errMsg = new Set();
+		const errDetails = new Set();
+		let errorKind: CodemodExecutionErrorType = 'errorRunningCodemod';
+
+		childProcess.stderr.on('data', function (err: Buffer) {
```
grzpab, line 277: `error: unknown`

src/components/engineService.ts
```diff
@@ -261,10 +270,49 @@ export class EngineService {
...
+					errDetails.add(error.caseTitle);
+				}
+				if ('group' in error) {
+					errDetails.add(error.group);
+				}
+				if ('kind' in error) {
+					errorKind = error.kind as CodemodExecutionErrorType;
+				}
+			} catch (e) {
+				errMsg.add(err.toString());
```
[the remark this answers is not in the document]
henoktsegayetoptal: we might run into an issue if the buffer couldn't be parsed that is usually the case if `JSON.parse` doesn't work

src/components/engineService.ts
```diff
@@ -261,10 +270,51 @@ export class EngineService {
...
-		childProcess.stderr.on('data', (data) => {
-			console.error(data.toString());
+		const errorMessage = new Set();
+		const errDetails = new Set();
+		let errorKind: CodemodExecutionErrorType = 'errorRunningCodemod';
+
+		childProcess.stderr.on('data', function (err: unknown) {
+			if (!(err instanceof Buffer)) return;
+			try {
+				const msg = err.toString();
```
grzpab, line 280: whatever happens here, it should create just one object

PR #196, opened 2023-03-29
Feat/int 690 show error messages from the codemod execution

henoktsegayetoptal:


src/components/engineService.ts
```diff
@@ -261,10 +268,42 @@ export class EngineService {
 			shell: true,
 		});
 
-		childProcess.stderr.on('data', (data) => {
-			console.error(data.toString());
+		const errorMessage = new Set<string>();
```
grzpab, line 271: `errorMessages`, also it can be just a mutable array

src/components/engineService.ts
```diff
@@ -261,10 +268,42 @@ export class EngineService {
...
 
+		childProcess.stderr.on('end', () => {
+			if (!errorMessage.size && !this.#execution?.affectedFiles.size) {
+				window.showWarningMessage(Messages.noAffectedFiles);
+			}
+
+			errorMessage.forEach((error) => {
+				try {
+					const parsedError = JSON.parse(error);
+					console.log('parsed error');
```
grzpab, line 291: you can remove it

src/components/engineService.ts
henoktsegayetoptal: since the backend sometimes emits the same message we want to make sure we have unique strings

PR #197, opened 2023-03-29
Feat/int 695 list available codemods by analyzing package.json

henoktsegayetoptal:


src/components/intuitaTreeDataProvider.ts
```diff
@@ -51,7 +52,9 @@ const getElementIconBaseName = (kind: Element['kind']): string => {
 	}
 };
 
-export class IntuitaTreeDataProvider implements TreeDataProvider<ElementHash> {
+export class IntuitaTreeDataProvider
+	implements TreeDataProvider<ElementHash | CodemodItem>
```
grzpab, line 56: `ElementHash` is a string that allows us to address a particular element in the tree using just its ID, codemod item should also generate its own hash digest. Please check the `Element` implementation for inspiration.

PR #205, opened 2023-03-30
Feat/int 695 list available codemods by analyzing package.json

henoktsegayetoptal:


grzpab:
@henoktsegayetoptal what do you think about separating these tree views into two distinct views? one for possible codemods and one for jobs/cases?

henoktsegayetoptal:
> @henoktsegayetoptal what do you think about separating these tree views into two distinct views? one for possible codemods and one for jobs/cases?

So the Idea was to have a fallback UI that shows a list of codemods when there is no job available, that was what we talked with alex

src/extension.ts
```diff
@@ -658,6 +674,15 @@ export async function activate(context: vscode.ExtensionContext) {
 			});
 		},
 	);
+	vscode.commands.registerCommand(
+		'intuita.runCodemod',
+		async (item: ElementHash) => {
```
grzpab, line 679: this should be probably called `CodemodHash` or `CodemodHashDigest`

src/extension.ts
```diff
@@ -74,7 +76,12 @@ export async function activate(context: vscode.ExtensionContext) {
 	);
 
 	new FileService(messageBus);
-
+	const rootPath =
+		vscode.workspace.workspaceFolders &&
+		vscode.workspace.workspaceFolders.length > 0
+			? vscode.workspace.workspaceFolders[0]?.uri.fsPath
```
grzpab, line 82: does the following line work?
``` vscode.workspace.workspaceFolders?.[0]?.uri.fsPath```

File: src/elements/CodemodList.ts [hunk 3]
```diff
@@ -0,0 +1,239 @@
...
+		leastVersionSupported: '4.0.0',
+		latestVersionSupported: '4.0.0',
+		leastSupportedUpgrade: '3.0.0',
+	},
+];
+
+const buildCodemodItemHash = (codemodItem: CodemodItem) => {
+	return buildHash(
+		`${codemodItem.label} ${codemodItem.id} ${codemodItem.commandToExecute}`,
+	) as ElementHash;
```
grzpab, line 98: you can introduce the `CodemodHash` type as well

src/elements/CodemodList.ts
```diff
@@ -0,0 +1,239 @@
...
+		'bluelightbulb.svg',
+	);
+}
+
+class CodemodTreeProvider {
+	rootPath: string | null | undefined;
+	#messageBus: MessageBus;
+	#codemodItemsMap: Map<ElementHash, CodemodItem> = new Map();
+
+	constructor(path: string | null | undefined, messageBus: MessageBus) {
```
grzpab, line 130: `string | null` IMHO, based on the observation that:
* `null` means that the `path` is set not to contain any value on purpose
* `undefined` means the `path` is undefined but you don't know why that is

src/elements/CodemodList.ts
```diff
@@ -0,0 +1,239 @@
...
+		}
+		return [];
+	}
+
+	public runCodemod(codemod: string) {
+		commands.executeCommand(codemod);
+	}
+
+	getTreeItem(element: ElementHash): CodemodItem {
+		const el = this.#codemodItemsMap.get(element) as CodemodItem;
```
grzpab, line 157: you can just return in the same line

src/elements/CodemodList.ts
```diff
@@ -0,0 +1,239 @@
...
+		return el;
+	}
+
+	private getDepsInPackageJson(): Map<ElementHash, CodemodItem> | null {
+		if (!this.rootPath) return null;
```
grzpab, line 162: `{ return null; }`

src/elements/CodemodList.ts
```diff
@@ -0,0 +1,239 @@
...
+		const packageJsonPath = path.join(this.rootPath, 'package.json');
+		if (this.pathExists(packageJsonPath)) {
```
grzpab, line 164: you can invert this whole statement by checking if the path does not exist.

src/elements/CodemodList.ts
```diff
@@ -0,0 +1,239 @@
...
+		dependencyName: string,
+		version: string,
+	): null | PackageUpgradeItem[] {
+		// replace ^, ~ , *
+		const actualVersion = version.replace(/[^0-9.]/g, '');
+
+		const codemod = packageUpgradeList.filter(
+			(el) => el.packageName === dependencyName,
+		);
+		if (!codemod) return null;
```
grzpab, line 221: `{ return null; }`

src/extension.ts
henoktsegayetoptal: Yes

grzpab (approved):

PR #207, opened 2023-03-31
Feature/int 709 show link to created issue in vscode

DmytroHryshyn:


src/components/webview/sourceControl.ts
```diff
@@ -9,6 +9,11 @@ interface UserAccountStorage {
 	getUserAccount(): string | null;
 }
 
+type CreateIssueResponse = {
+	url: string;
```
grzpab, line 13: what's the difference between `url` and `html_url`?

src/extension.ts
```diff
@@ -203,10 +203,18 @@ export async function activate(context: vscode.ExtensionContext) {
...
 					if (decoded._tag === 'Right') {
-						await sourceControl.createIssue(decoded.right);
-						vscode.window.showInformationMessage(
-							'Successfully created issue.',
+						const { html_url } = await sourceControl.createIssue(
+							decoded.right,
 						);
+						const messageSelection =
+							await vscode.window.showInformationMessage(
+								'Successfully created issue',
```
grzpab, line 211: do you want to show the link URL here as well?

src/components/webview/sourceControl.ts
DmytroHryshyn: `url` is the path of GET api endpoint ("https://api.github.com/repos/DmytroHryshyn/test_repo/issues/31"), and `html_ url` is normal url. I think we dont need `url` at all, i will remove it

src/extension.ts
DmytroHryshyn: I think i will be nice to have it

grzpab (approved):

PR #209, opened 2023-03-31
Feature/int 711 when job or case is clicked show create

DmytroHryshyn:


src/components/messageBus.ts
```diff
@@ -51,21 +51,21 @@ export const enum MessageKind {
 	/**
 	 * account events
 	 */
-	onAfterLinkedAccount = 26,
-	onAfterUnlinkedAccount = 27,
+	afterLinkedAccount = 26,
```
grzpab, line 54: how about `accountLinked`

src/components/messageBus.ts
```diff
@@ -51,21 +51,21 @@ export const enum MessageKind {
...
+	afterUnlinkedAccount = 27,
```
grzpab, line 55: how about `accountUnlinked`?

src/components/messageBus.ts
```diff
@@ -51,21 +51,21 @@ export const enum MessageKind {
...
 
 	/**
 	 * config events
 	 */
 
-	onAfterConfigurationChanged = 28,
+	afterConfigurationChanged = 28,
```
grzpab, line 61: how about `configurationChanged`?

src/components/messageBus.ts
```diff
@@ -51,21 +51,21 @@ export const enum MessageKind {
...
 
 	/**
 	 * source control
 	 */
 
-	onBeforeCreateIssue = 29,
-	onAfterCreateIssue = 30,
+	beforeCreateIssue = 29,
```
grzpab, line 67: how about `beforeIssueCreated`

src/components/messageBus.ts
```diff
@@ -51,21 +51,21 @@ export const enum MessageKind {
...
+	afterCreateIssue = 30,
```
grzpab, line 68: how about `afterIssueCreated`?

src/extension.ts
```diff
@@ -180,17 +180,17 @@ export async function activate(context: vscode.ExtensionContext) {
...
+			const panelInstance = IntuitaPanel.getInstance(
 				context,
 				{ getConfiguration },
 				globalStateAccountStorage,
 				messageBus,
 			);
-
-			instance.render();
+			await panelInstance.render();
+			panelInstance.setFormData({ title: treeItem.label as string });
```
grzpab, line 193: what's wrong with the `treeItem.label`, isn't it a `string`?

src/components/webview/IntuitaPanel.ts
```diff
@@ -145,8 +249,27 @@ export class IntuitaPanel {
 			return;
 		}
 
-		this.__view.onDidReceiveMessage((message: WebViewMessage) => {
-			commands.executeCommand(message.command, message.value);
+		this.__view.onDidReceiveMessage((message: WebviewResponse) => {
```
grzpab, line 252: I'd separate the callback from the function, this way it's easier to read it.

src/components/webview/IntuitaPanel.ts
```diff
@@ -89,7 +125,41 @@ export class IntuitaPanel {
 	}
 
 	public render() {
-		this.__panel?.reveal();
+		const initWebviewPromise = new Promise((resolve, reject) => {
+			this.__panel?.reveal();
+
+			const timeout = setTimeout(() => {
+				this.__panel?.dispose();
```
grzpab, line 132: is it a promise?

DmytroHryshyn: panel.reveal is sync method, but it takes some time to mount react application (and subscribe to messages) after its revealed. And because of that there can be glitches where webview ignores some initial events. So i had to wrap it in a promise. I will rewrite this method if i find better solution.

grzpab (approved):

PR #213, opened 2023-04-04
[IDE ext] Re-executing recipe after dismissing case/job in the previous execution ignores the case/job

hbjORbj:
Fixes https://linear.app/intuita/issue/INT-732/[ide-ext]-re-executing-recipe-after-dismissing-casejob-in-the-previous#comment-135c13bd

- Tested
- `yarn package` succeeds

### DIFF
--- a/src/components/jobManager.ts
+++ b/src/components/jobManager.ts
@@ -19,16 +19,10 @@ export class JobManager {
 	readonly #messageBus: MessageBus;
 
 	#jobMap: Map<JobHash, Job>;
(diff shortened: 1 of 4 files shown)


grzpab (commented):
@hbjORbj can you remove all mentions of `rejectedJobHashes` from the codebase? I think it's in the job manager and in serialization services.

hbjORbj:
Sure! Done.

grzpab (approved):

PR #214, opened 2023-04-04
Feat/int 717 automatically detect git remote path in vscode

henoktsegayetoptal:
---

<details open="true"><summary>Generated summary (powered by <a href="https://app.graphite.dev">Graphite</a>)</summary>

> #### What Changed
> Added
> - DefaultConfigurationClass
> - extensionDependencies
> #### Testing
> Test that the extension works as expected by activating it and checking that the repository path is set correctly. Also, check that the extension dependencies are installed.
> #### Reasoning
> The DefaultConfigurationClass was added to handle the configuration of the extension, and extensionDependencies were added to ensure that the required dependencies are installed for the extension to work properly.
</details>


src/configuration.ts
```diff
@@ -41,4 +43,76 @@ export const getConfiguration = () => {
...
+								'repositoryPath',
+								repositoryPath,
+								vscode.ConfigurationTarget.Global,
+							);
+					}
+				} catch (error) {
+					if (error instanceof Error) {
+						vscode.window.showErrorMessage(error.message);
+					}
+					console.error('error', error);
```
grzpab, line 72: You can just write `console.error(error)`

src/configuration.ts
```diff
@@ -41,4 +43,76 @@ export const getConfiguration = () => {
...
+		this.#messageBus.subscribe(MessageKind.extensionActivated, async () => {
+			const repositoryPathFromConfig = getConfiguration().repositoryPath;
+			if (!repositoryPathFromConfig) {
+				try {
+					const repositoryPath = await Promise.race([
+						this.getGitRemote(),
+						this.timeout(4000),
+					]);
+
+					if (repositoryPath) {
```
grzpab, line 59: I'd revert the logic here (so return if `!repositoryPath`)

src/configuration.ts
```diff
@@ -41,4 +43,76 @@ export const getConfiguration = () => {
...
+						vscode.window.showErrorMessage(error.message);
+					}
+					console.error('error', error);
+				}
+			}
+		});
+	}
+	getGitRemote = () =>
+		// eslint-disable-next-line no-async-promise-executor
+		new Promise(async (resolve, reject) => {
```
grzpab, line 79: actually I think it's a serious antipattern: https://stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it/23803744#23803744

I think you can make it an async function and create the promise within it when needed.

src/configuration.ts
```diff
@@ -41,4 +43,76 @@ export const getConfiguration = () => {
...
+			const gitExtension = vscode.extensions.getExtension(
```
grzpab, line 80: can it throw?

grzpab (changes requested):
See inline comments

src/configuration.ts
henoktsegayetoptal: it can reject , not throw

grzpab (changes requested):
See inline comments

src/configuration.ts
```diff
@@ -41,4 +44,74 @@ export const getConfiguration = () => {
...
+
+			if (!gitExtension) {
+				return reject(new Error('Git extension not found'));
+			}
+			if (!gitExtension.isActive) {
+				gitExtension
+					.activate()
+					.then(() =>
+						this.getPathFromGitExtension(gitExtension, resolve),
+					);
```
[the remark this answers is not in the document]
henoktsegayetoptal: unfortunately `activate()` doesn't have a catch

src/types/git.d.ts
```diff
@@ -0,0 +1,209 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
```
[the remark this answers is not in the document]
grzpab: did @DmytroHryshyn add the same file already?

grzpab (approved):

PR #216, opened 2023-04-05
Feature/int 736 update apply flow

DmytroHryshyn:


src/cases/caseManager.ts
```diff
@@ -29,9 +29,10 @@ export class CaseManager {
 		this.#messageBus.subscribe(MessageKind.acceptCase, (message) =>
 			this.#onAcceptCaseMessage(message),
 		);
-		this.#messageBus.subscribe(MessageKind.jobsAccepted, (message) =>
-			this.#onJobsAcceptedOrJobsRejectedMessage(message),
-		);
+		// We should not delete the job after "apply". User may want to create PR for this job
```
grzpab, line 32: Then I suppose we need to still send this event and keep the job, but not display it? Actually we should keep it as an accepted job with some UI tag telling its state?

git.d.ts
```diff
@@ -0,0 +1,390 @@
+/*---------------------------------------------------------------------------------------------
```
grzpab, line 1: @henoktsegayetoptal I think you also worked on these git definitions?

src/components/jobManager.ts
```diff
@@ -228,16 +228,16 @@ export class JobManager {
...
-				const otherJobHashes =
-					this.#uriHashJobHashSetManager.getRightHashesByLeftHash(
-						uriHash,
-					);
-
-				for (const jobHash of otherJobHashes) {
-					this.#uriHashJobHashSetManager.delete(uriHash, jobHash);
-					this.#jobMap.delete(jobHash);
-				}
+				// We should not delete the job after "apply". User may want to create PR for this job
```
grzpab, line 231: I am okay with removing these statements, but how would these jobs be visible in the UI?

src/components/webview/repository.ts
```diff
@@ -0,0 +1,84 @@
+import { APIState } from '../../../git';
+import { API, Repository } from '../../../git';
+import { assertsNeitherNullOrUndefined } from '../../utilities';
+
+const branchNameFromStr = (str: string): string => {
+	let branchName = str.toLowerCase();
```
grzpab, line 6: I think you can chain the following 4 statements.

src/components/webview/repository.ts
```diff
@@ -0,0 +1,84 @@
...
+export class UninitializedError extends Error {}
+
+export class RepositoryService {
+	__repo: Repository | null = null;
+
+	constructor(private readonly __gitAPI: API) {
+		this.__gitAPI.onDidChangeState(this.onDidChangeState);
+	}
+
+	private onDidChangeState = (state: APIState) => {
```
grzpab, line 36: I would start the private methods with `__` (double underscore) for better visibility 😉

src/extension.ts
```diff
@@ -186,6 +190,16 @@ export async function activate(context: vscode.ExtensionContext) {
 		messageBus,
 	);
 
+	const gitExtension =
+		vscode.extensions.getExtension<GitExtension>('vscode.git');
```
grzpab, line 194: @henoktsegayetoptal seems to me you and @rajasegar need to cooperate on both of your PRs so you don't merge conflicting changes.

src/extension.ts
```diff
@@ -210,29 +224,82 @@ export async function activate(context: vscode.ExtensionContext) {
...
+				}
+
+				const hasChanges =
+					await repositoryService.hasWorkingTreeChanges();
+
+				if (!hasChanges) {
+					throw new Error('Nothing to commit');
+				}
+
+				const jobHash = typeof arg0 === 'string' ? arg0 : null;
```
grzpab, line 247: I would perform this operation first in the command handler.

src/extension.ts
```diff
@@ -210,29 +224,82 @@ export async function activate(context: vscode.ExtensionContext) {
...
+
+				const targetBranch = repositoryService.getBranchName(
+					jobHash,
+					jobTitle,
+				);
+
+				await panelInstance.render();
+
+				// @TODO check why branch name can be undefined
+				const currentBranchName = currentBranch.name ?? '';
```
grzpab, line 282: A short explaination is when you're on (HEADed) a tag or the commit (detached HEAD)

src/components/webview/repository.ts
```diff
@@ -0,0 +1,84 @@
...
+		if (state === 'initialized') {
+			this.__repo = this.__gitAPI.repositories[0] ?? null;
+		}
+	};
+
+	public getAllBranches = async () => {
```
grzpab, line 42: I would not use arrow functions as methods, this is syntax sugar that is confusing (at least to me).

src/components/webview/repository.ts
```diff
@@ -0,0 +1,84 @@
...
+
+		return changes.length !== 0;
+	};
+
+	public getBranchName = (jobHash: string, jobTitle: string) => {
+		return branchNameFromStr(`${jobTitle}-${jobHash}`);
+	};
+
+	public getBaseBranchName = () => {
+		return 'main';
```
grzpab, line 73: how do we know that?

grzpab (changes requested):
See inline comments

src/components/webview/repository.ts
DmytroHryshyn: @grzpab this method is not used, i always branch from current branch. I forgot to remove it.

src/extension.ts
DmytroHryshyn: @grzpab i've checked Henok's PR, we dont have conflicting changes, we just access the vscode.git from different places. But it will be probably better to move remote detection logic to the RepositoryService.

git.d.ts
henoktsegayetoptal: yep I removed the git error codes but yes

src/cases/caseManager.ts
DmytroHryshyn: @grzpab Yes, i think we need to keep it in the tree, but remove apply/dismiss buttons in accepted state, leaving only one option - create PR. Do i need to implement this in this PR?

src/components/jobManager.ts
DmytroHryshyn: I will probably add third type of tree items for accepted jobs, and it will contain only Create PR button

src/cases/caseManager.ts
DmytroHryshyn: Added "accepted" state to the current PR

src/components/jobManager.ts
```diff
@@ -228,16 +250,16 @@ export class JobManager {
...
```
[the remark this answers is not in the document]
DmytroHryshyn: @grzpab But as i understood, onJobsAcceptedOrJobsRejectedMessage removes all case jobs and the case itself. But we want to keep both jobs and case. So we should ignore this event. Did i missed something? 

Removed comments.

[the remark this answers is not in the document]
grzpab: True, good point.

grzpab (approved):

PR #217, opened 2023-04-05
[IDE ext] Apply change from one recipe shouldn't affect change from other recipes even if they are applying to the same file

hbjORbj:
https://linear.app/intuita/issue/INT-722/[ide-ext]-apply-change-from-one-recipe-shouldnt-affect-change-from#comment-7eacaa56

### Before

https://user-images.githubusercontent.com/32841130/230227257-075d538d-45a0-40eb-86b0-9e1aa6124e3a.mov

### After

https://user-images.githubusercontent.com/32841130/230227213-910d5c7f-997d-4df8-a94d-bcca8db58273.mov

### DIFF
--- a/src/components/jobManager.ts
+++ b/src/components/jobManager.ts
@@ -236,7 +236,6 @@ export class JobManager {
 
 				for (const jobHash of otherJobHashes) {
 					this.#uriHashJobHashSetManager.delete(uriHash, jobHash);
-					this.#jobMap.delete(jobHash);
 				}
 			}
 


src/components/jobManager.ts
```diff
@@ -236,7 +236,6 @@ export class JobManager {
 
 				for (const jobHash of otherJobHashes) {
 					this.#uriHashJobHashSetManager.delete(uriHash, jobHash);
-					this.#jobMap.delete(jobHash);
```
grzpab, line 239: I would say delete only if there are no other uriHashes that depend on that job.

grzpab (changes requested):
See inline comments

src/components/jobManager.ts
hbjORbj: Umm, I see that this whole for loop is removed in `main`.

hbjORbj:
Closing this PR since https://github.com/intuita-inc/intuita-vscode-extension/pull/216 already achieved the same goal.

PR #225, opened 2023-04-07
remove source control view, fix base url

DmytroHryshyn:


src/components/webview/sourceControl.ts
```diff
@@ -48,7 +48,7 @@ export class SourceControlService {
 		this.__messageBus.publish({ kind: MessageKind.beforePRCreated });
 
 		const result = await axios.post<CreatePRResponse>(
-			'https://telemetry.intuita.io/sourceControl/github/pulls',
+			'https://codemod-studio.vercel.app/sourceControl/github/pulls',
```
grzpab, line 51: I think this was correct, am I right?

src/components/webview/sourceControl.ts
```diff
@@ -82,7 +82,7 @@ export class SourceControlService {
 		this.__messageBus.publish({ kind: MessageKind.beforeIssueCreated });
 
 		const result = await axios.post<CreateIssueResponse>(
-			'https://telemetry.intuita.io/sourceControl/github/issues',
+			'https://codemod-studio.vercel.app/sourceControl/github/issues',
```
grzpab, line 85: same here?

src/components/webview/sourceControl.ts
DmytroHryshyn: i've already reverted that change. At first i thought that we moved our backend to the new host

grzpab (approved):

PR #228, opened 2023-04-07
[IDE ext.] Selecting Apply/Discard in Intuita tab leads to Explorer Tab

hbjORbj:
https://linear.app/intuita/issue/INT-756/[ide-ext]-selecting-applydiscard-in-intuita-tab-leads-to-explorer-tab

### Before

https://user-images.githubusercontent.com/32841130/230590076-666b6728-5503-44f0-bd3d-3033f3b4a9ce.mov

### After

https://user-images.githubusercontent.com/32841130/230590004-73056523-8920-446b-a953-6bec7e91796a.mov

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -121,6 +121,14 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	treeDataProvider.setReveal(explorerTreeView.reveal);
 
+	explorerTreeView.onDidChangeVisibility(({ visible }) => {
+		if (visible) {
+			treeDataProvider.setReveal(explorerTreeView.reveal);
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -604,6 +604,14 @@ export async function activate(context: vscode.ExtensionContext) {
 		]);
 	};
 
+	explorerTreeView.onDidChangeVisibility(({ visible }) => {
```
grzpab, line 607: can you move it closer to the `.reveal()` instruction in this file?

hbjORbj: addressed, merging PR

PR #229, opened 2023-04-07
allow to reject the job and case in accepted state

DmytroHryshyn:


src/components/fileService.ts
```diff
@@ -64,7 +52,7 @@ export class FileService {
 
 		this.#messageBus.publish({
 			kind: MessageKind.deleteFiles,
-			uris: [message.oldUri, message.newContentUri],
```
grzpab, line 67: what does this change do?

src/components/jobManager.ts
```diff
@@ -330,7 +330,10 @@ export class JobManager {
 			if (
 				job &&
 				(job.kind === JobKind.rewriteFile ||
-					job.kind === JobKind.moveAndRewriteFile) &&
+					job.kind === JobKind.moveAndRewriteFile ||
```
grzpab, line 333: can you explain this change?

src/components/fileService.ts
DmytroHryshyn: We dont want to remove codemod output files from storage immediately after changes are applied, so i've removed message.newContentUri from that array.

src/components/jobManager.ts
DmytroHryshyn: This should clean all output files of the job from storage when the job is rejected

grzpab (approved):

PR #232, opened 2023-04-07
Feat/int 755 unable to detect packagejson in mono repo and others

henoktsegayetoptal:


henoktsegayetoptal:
Moved to https://github.com/intuita-inc/intuita-vscode-extension/pull/233

PR #233, opened 2023-04-10
Feat/int 765 support running codemods from the path and sort

henoktsegayetoptal:


henoktsegayetoptal:
Screenshot 2023-04-10 at 3 18 34 PM

packageJsonAnalyzer/types.ts
```diff
@@ -0,0 +1,22 @@
+export type CodemodHash = string & { __type: 'CodemodHash' };
+export type PathHash = string & { __type: 'PathHash' };
+
+export type Path = {
```
grzpab, line 4: what's the reason for this type?

src/extension.ts
```diff
@@ -686,15 +515,15 @@ export async function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(
 		vscode.commands.registerCommand(
 			'intuita.executeNextJsCodemods',
-			async () => {
+			async (path?: vscode.Uri) => {
```
grzpab, line 518: I think the `path` should be sent obligatorily here from now on.

The reason is that we can delegate figuring out the correct URI (path) to the caller of this command. This way the code gets less branch-y in the command and the repetition is lower.

(This is BTW one of the reasons why I don't want to have optional parameters, they always means branching down the line).

packageJsonAnalyzer/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+			) {
+				return el;
+			}
+
+			return null;
+		})
+		.filter(isNeitherNullNorUndefined);
+};
+
+export const pathExists = (p: string): boolean => {
```
grzpab, line 44: `p` - `path`

packageJsonAnalyzer/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+	try {
+		accessSync(p);
```
grzpab, line 46: can you check if there's a workspace API that can ensure that the path exists?Ideally you shouldn't use the file system API unless necessary.

packageJsonAnalyzer/codemodItem.ts
```diff
@@ -0,0 +1,31 @@
+import { TreeItem, TreeItemCollapsibleState } from 'vscode';
+import path from 'path';
+import { buildCodemodItemHash } from './utils';
+import { CodemodHash } from './types';
+
+export class CodemodItem extends TreeItem {
```
grzpab, line 6: I think this should be a separate POJSO type (like ElementItem) and there should be a builder that maps that type into a tree item.

The rationale is that here the new derived class just overrides that properties of `TreeItem`, there are no new specific methods to that class.

packageJsonAnalyzer/codemodList.ts
```diff
@@ -0,0 +1,275 @@
...
+import { watchFileWithPattern } from '../src/fileWatcher';
+import { CodemodHash, PathHash, Path, PackageUpgradeItem } from './types';
+import { CodemodItem } from './codemodItem';
+import { commandList } from './constants';
+import { checkIfCodemodIsAvailable, pathExists } from './utils';
+
+class CodemodTreeProvider {
+	#rootPath: string | null;
+	#messageBus: MessageBus;
+	#codemodItemsMap: Map<CodemodHash | PathHash, CodemodItem | Path> =
```
grzpab, line 24: what is this map responsible for? Is it mapping the hash digests of tree items to the very tree items?

packageJsonAnalyzer/codemodList.ts
```diff
@@ -0,0 +1,275 @@
...
+	public readonly onDidChangeTreeData: Event<void>;
+
+	constructor(rootPath: string | null, messageBus: MessageBus) {
+		this.#rootPath = rootPath;
+		this.#messageBus = messageBus;
+		this.onDidChangeTreeData = this.#eventEmitter.event;
+		this.getPackageJsonListAndWatch();
+
+		this.#messageBus.subscribe(MessageKind.runCodemod, (message) => {
+			const codemodItemFound = this.getTreeItem(
```
grzpab, line 36: should we display a message that if the codemod item has been found, we are aborting the "same" execution instance?

packageJsonAnalyzer/codemodList.ts
henoktsegayetoptal: yes

packageJsonAnalyzer/codemodList.ts
henoktsegayetoptal: what do you mean "aborting the "same" execution instance?"

packageJsonAnalyzer/codemodList.ts
```diff
@@ -0,0 +1,275 @@
...
+			const pathFromRoot = uri.fsPath
+				.replace('/package.json', '')
+				.replace(this.#rootPath as string, '');
+
+			const splitParts = pathFromRoot.split('/');
+			CodemodsFromPackageJson.forEach((codemodItem, codemodHash) => {
+				codemods.set(codemodHash, codemodItem);
+			});
+			splitParts.forEach((part, index, parts) => {
+				const currentPWD = `${this.#rootPath}${parts
```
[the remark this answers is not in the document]
henoktsegayetoptal: it was meant to be `WD` - working directory

packageJsonAnalyzer/codemodList.ts
grzpab: this snippet won't launch another execution of the same command if it's been run, am I right?

packageJsonAnalyzer/types.ts
henoktsegayetoptal: that is the path type that includes the codemodes

packageJsonAnalyzer/codemodList.ts
henoktsegayetoptal: That is a good point but right now it doesn't have a restriction.

packageJsonAnalyzer/utils.ts
henoktsegayetoptal: we can use the async methods

grzpab (approved):

PR #235, opened 2023-04-11
fix issue with panels

DmytroHryshyn:


src/components/webview/SourceControlWebviewPanel.ts
```diff
@@ -5,36 +5,35 @@ import {
...
 
 	static getInstance(
 		context: ExtensionContext,
 		messageBus: MessageBus,
 		repositoryService: RepositoryService,
 		userAccountStorage: UserAccountStorage,
 	) {
-		if (this.__instance) {
-			return this.__instance;
+		if (!SourceControlWebviewPanel.instance) {
```
grzpab, line 27: is it possible we don't have a static property at all? so we basically end up with a proper dependency injection mechanism?

DmytroHryshyn: this will be complicated. panel should follow singleton pattern and should be recreated only if it was disposed. otherwise we will see multiple panels.

grzpab (approved):

PR #236, opened 2023-04-12
fix webpack dev mode

DmytroHryshyn:
- adds ability to run apps in dev mode
- replace rewire with craco


intuita-webview/craco.config.ts
```diff
@@ -0,0 +1,55 @@
...
+	path.resolve(appDirectory, relativePath);
+
+module.exports = {
+	webpack: {
+		configure: (
+			config: Configuration,
+			{ env }: { env: 'production' | 'development' },
+		) => {
+			if (env === 'production') {
+				if (config.optimization) {
```
grzpab, line 16: should we just initialize the `optimization` object if it's not there?

henoktsegaye (approved):

grzpab (approved):

PR #238, opened 2023-04-12
fix

DmytroHryshyn:


src/components/webview/MainWebviewProvider.ts
```diff
@@ -480,22 +480,34 @@ export class IntuitaProvider implements WebviewViewProvider {
...
-					right,
+					leftUri,
+					rightUri,
 					title,
 				);
 			}
 
+			if (message.value.command === 'vscode.open') {
+				const args = message.value.arguments;
+				if (!args?.[0]?.path) {
```
grzpab, line 503: does `args[0] instanceof Uri` work?

DmytroHryshyn: no, it will be false because we pass it from webview. I think it serializes it to string and then back when we use postMessage. (i will double check that)

grzpab (approved):

PR #240, opened 2023-04-12
fix

DmytroHryshyn:


src/components/jobManager.ts
```diff
@@ -346,6 +346,8 @@ export class JobManager {
 			this.#jobMap.delete(jobHash);
 		}
 
+		messages.push({ kind: MessageKind.updateElements });
```
grzpab, line 349: was it accidentally removed?

DmytroHryshyn: no, this line was missing from the beginning, just checked the git timeline

grzpab (approved):

PR #244, opened 2023-04-13
fix pr flow, fix codicons

DmytroHryshyn:


resources/codicon.css
```diff
@@ -0,0 +1,474 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
```
grzpab, line 3: Can you remove the `See License.txt in the project root for license information.`
and add the full text of the MIT License in this comment?

grzpab: Also, can you add the link you sourced that file from?

grzpab (approved):

PR #247, opened 2023-04-14
Apply color to TreeItem on focus and reduce indent in Tree

hbjORbj:
https://linear.app/intuita/issue/INT-797/apply-color-to-treeitem-on-focus-and-reduce-indent-in-tree

Before and After videos are inside the linear task above.


intuita-webview/src/main/TreeView/index.tsx
```diff
@@ -83,11 +94,13 @@ const TreeView = ({ node }: Props) => {
 				id={node.id}
 				label={node.label ?? ''}
 				icon={icon}
-				indent={depth * 30}
+				indent={depth * 12}
```
grzpab, line 97: what is the unit of it? `px`? `rem`?

hbjORbj: in px. We do `<div style={{ minWidth: `${indent}px` }} />` when building `TreeItem`.

grzpab:
@hbjORbj actually I think that the initial "Recipe" node has too much whitespace on its left "bank", the rest of the icon placement is good. We could actually switch the folder icon with the one available in the vscode directly.

hbjORbj:
Addressed your comments. I add an icon for the ROOT kind as well. I think this icon makes sense because recipe does the fixes and the patches.

Screenshot 2023-04-14 at 09 06 18

henoktsegayetoptal (approved):

intuita-webview/src/assets/wrench.svg
grzpab, line 1: who's the creator of that asset?

hbjORbj: answered in slack

PR #252, opened 2023-04-14
Hotfix/fix panel names

DmytroHryshyn:


src/components/webview/WebviewResolver.ts
```diff
@@ -12,6 +12,7 @@ export class WebviewResolver {
 				Uri.joinPath(this.__extensionPath, 'intuita-webview/build'),
 				Uri.joinPath(this.__extensionPath, 'resources'),
 			],
+			retainContextWhenHidden: true,
```
grzpab, line 15: what does that do?

src/components/webview/WebviewPanel.ts
```diff
@@ -0,0 +1,120 @@
+import { Disposable, Uri, ViewColumn, WebviewPanel, window } from 'vscode';
+import { WebviewResolver } from './WebviewResolver';
+import { WebviewMessage } from './webviewEvents';
+import { Message, MessageBus, MessageKind } from '../messageBus';
+
+type PanelOptions = Readonly<{
+	type: string;
+	title: string;
+	showOptions:
```
grzpab, line 9: IMHO the fewer possibilities of states, the better.

I would not have `showOptions` attribute at all, but a `viewColumn` attribute and `preserveFocus: boolean`

src/components/webview/WebviewPanel.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export type Options = PanelOptions & WebviewOptions;
+
+export abstract class IntuitaWebviewPanel {
+	protected __panel: WebviewPanel | null = null;
+	protected __disposables: Disposable[] = [];
+	protected __webviewMounted = false;
+
+	constructor(
+		protected readonly options: Options,
```
grzpab, line 31: Can you make it `_options`?

src/components/webview/WebviewPanel.ts
```diff
@@ -0,0 +1,120 @@
...
+type WebviewOptions = Readonly<{
+	extensionUri: Uri;
+	webviewName: string;
+	initialData: unknown;
+}>;
+
+export type Options = PanelOptions & WebviewOptions;
+
+export abstract class IntuitaWebviewPanel {
+	protected __panel: WebviewPanel | null = null;
```
grzpab, line 26: if it's protected, let's perhaps make it `_panel`

`x` => public
`_x` => protected
`__x` => private

does that make sense (it's loosely following the Python conventions)?

src/components/webview/SourceControlWebviewPanel.ts
```diff
@@ -1,95 +1,24 @@
...
-						disposable?.dispose();
-						this.__webviewMounted = true;
-						resolve(null);
-					}
-				},
-			);
-		});
-
-		return initWebviewPromise;
+	public dispose() {
```
grzpab, line 19: can you try to add `override` (it's a new TS keyword) and see if it complains should you remove `super.dispose()`?

src/components/webview/WebviewResolver.ts
DmytroHryshyn: this keeps webview active when we switch to other tabs, if set to false webview will unmount each time

src/components/webview/WebviewPanel.ts
DmytroHryshyn: in js its usually `_x` for private and `x` for everything else.

DmytroHryshyn: we definitely need to add this to our code style guide doc

grzpab (approved):

PR #255, opened 2023-04-17
Feature/int 812 persist remote selected by user

DmytroHryshyn:


src/persistedState/persistedStateService.ts
```diff
@@ -13,21 +14,23 @@ export class PersistedStateService {
...
 			debouncedOnUpdateElementsMessage(),
 		);
+
 		this.messageBus.subscribe(MessageKind.clearState, () =>
 			this.#onClearStateMessage(),
 		);
 	}
 
-	async #onUpdateElementsMessage() {
+	public async saveExtensionState() {
```
grzpab, line 33: why did you change the method name? actually it made sense to have it connected to the message kind

src/components/webview/webviewEvents.ts
```diff
@@ -106,6 +106,16 @@ export type WebviewResponse =
 	| Readonly<{
 			kind: 'webview.command';
 			value: Command;
+	  }>
+	| Readonly<{
+			kind: 'webview.createPR.afterFormDataChanged';
```
grzpab, line 111: can be just `webview.createPR.formDataChanged`

src/components/webview/repository.ts
```diff
@@ -27,31 +20,43 @@ const branchNameFromStr = (str: string): string => {
 	return branchName;
 };
 
+type PersistedState = {
```
grzpab, line 23: 1) I would not add `null` to an object type
2) can we get rid of the possibility of a nullable `PersistedState`?

src/persistedState/persistedStateService.ts
```diff
@@ -84,11 +90,13 @@ export class PersistedStateService {
 
 		const localStateUri = Uri.joinPath(uri, 'localState.json');
 
+		// @TODO add ability to rewrite state partially
```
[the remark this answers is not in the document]
DmytroHryshyn: Currently we cannot update specific field. We can only write the whole state to the file. I think ideally API should be something like
```saveState(key: string, value: unknown): void; getState("key"): unknown; clear(): void```

File: src/persistedState/persistedStateService.ts (hunk 1 above)
Comment (DmytroHryshyn) [reply src/persistedState/persistedStateService.ts:33]: I wanted to call this method directly so i made it public and changed the name to neutral. Should i add new events instead, and subscribe to them instead?

File: src/components/webview/repository.ts (hunk 3 above)
Comment (DmytroHryshyn) [reply src/components/webview/repository.ts:23]: getPersistedState reads from file system, we dont want to throw error when extension is loaded for the first time, so we return State | null, this makes sense. I can pass  just `remoteUrl ?? null ` instead of passing the whole state to repository service.

File: src/persistedState/getPersistedState.ts [hunk 5]
```diff
@@ -28,7 +32,10 @@ export const getPersistedState = async (
 		const str = buffer.toString();
 		const json = JSON.parse(str);
 
-		const persistedStateEither = persistedStateCodec.decode(json);
+		const persistedStateEither = persistedStateCodec.decode({
```
grzpab, line 35: oh I get what you wanted to do now :)

https://github.com/Effect-TS/schema#default-values we can use this instead (and temporarily, until people migrate to the new version of the persisted state)

src/components/webview/repository.ts
grzpab: Good question.

We can initialize properties with their "natural" neutral state (neutral as in the neutral value of on operation in the group theory):
- objects => null,
- strings => empty,
- numbers => 0
- arrays => empty list
- etc.

grzpab (approved):

PR #256, opened 2023-04-18
add ability to refresh jobs

DmytroHryshyn:


package.json
```diff
@@ -49,6 +49,10 @@
 			}
 		],
 		"commands": [
+			{
+				"command": "intuita.refreshStaleJobs",
```
grzpab, line 53: maybe make it just `refresh jobs`?

src/components/messageBus.ts
```diff
@@ -112,6 +112,13 @@ export type Command =
 			engine: Engine;
 			storageUri: Uri;
 			uri: Uri;
+	  }>
+	| Readonly<{
+			kind: 'executeCodemod';
```
grzpab, line 117: is that supported by the nora node engine?

src/components/jobManager.ts
```diff
@@ -406,4 +409,81 @@ export class JobManager {
...
+					engine: 'node',
+					storageUri,
+					uris: filePaths,
+				},
+				executionId,
+				happenedAt,
+			}),
+		);
+
+		console.log(messages);
```
grzpab, line 446: can you remove it?

src/components/jobManager.ts
```diff
@@ -406,4 +409,81 @@ export class JobManager {
...
+
+		// for (const message of messages) {
+		// 	this.#messageBus.publish(message);
+		// }
+	}
+
+	public async getStaleJobs(): Promise<Job[]> {
```
grzpab, line 453: shouldn't it be private?

src/components/engineService.ts
```diff
@@ -211,6 +214,30 @@ export class EngineService {
 				return args;
 			}
 
+			// execute single codemod
+			if (
+				'kind' in message.command &&
+				message.command.kind === 'executeCodemod'
+			) {
+				// @TODO engine should support such calls
```
grzpab, line 222: so this feature is not implemented yet right?

src/components/messageBus.ts
DmytroHryshyn: not yet,  that is what we need to add to the engine

src/components/jobManager.ts
DmytroHryshyn: I think sometimes we may want to get stale jobs to display them somewhere

src/components/engineService.ts
DmytroHryshyn: yes, its not implemented, but i still can work on this feature and connect it to the node engine when its done.

src/extension.ts
```diff
@@ -68,6 +68,11 @@ import { randomBytes } from 'crypto';
 const messageBus = new MessageBus();
 
 export async function activate(context: vscode.ExtensionContext) {
+
+	if(!context.storageUri) {
+		throw new Error('Storage Uri not found');
```
grzpab, line 73: when can this happen?

src/components/jobManager.ts
```diff
@@ -327,4 +333,45 @@ export class JobManager {
...
+				codemodHash,
+				engine: 'node',
+				storageUri: this.__storageUri,
+				uri,
+			},
+			executionId,
+			happenedAt,
+		}));
+
+		console.log(messages, 'test');
```
grzpab, line 371: can you remove it?

src/components/jobManager.ts
```diff
@@ -327,4 +333,45 @@ export class JobManager {
...
+			if(jobCodemod === null) {
+				return;
+			} 
+			codemodHashUriTuple.push([jobCodemod.hashDigest, job.oldUri]);
+		});
+
+		const executionId = buildExecutionId();
+		const happenedAt = String(Date.now());
+
+		const messages: Message[] = codemodHashUriTuple.map(([codemodHash, uri]) =>({
```
grzpab, line 358: how does that work with the principle that only one execution can be managed?

DmytroHryshyn: Currently we can have only 1 execution at a time, and we can only pass 1 codemod to the engine, right? 
I will need to add ability to `await` for execution to complete and then chain executions. 
Something like 
```
for(const codemod of codemods) {
await this.__engineService.executeCodemod(codemod, uris);
}
```

src/extension.ts
DmytroHryshyn: storageUri has type `uri | undefined`, but im not sure in what cases it can be undefined.

PR #258, opened 2023-04-18
Fix diff view bug

hbjORbj:


henoktsegayetoptal (approved):

intuita-webview/src/jobDiffView/DiffViewer/index.tsx
```diff
@@ -25,17 +25,18 @@ export const JobDiffView = ({
 			? 'inline'
 			: 'side-by-side';
 	});
-	const [isVisible, setVisible] = useState(false);
+	const [isVisible, setVisible] = useState(true);
```
grzpab, line 28: why did you need to invert the logic?

intuita-webview/src/jobDiffView/DiffViewer/index.tsx
```diff
@@ -25,17 +25,18 @@ export const JobDiffView = ({
...
-	const toggleViewed = useCallback(() => setVisible((v) => !v), [setVisible]);
-
-	useEffect(() => {
-		if (isVisible) {
-			collapsableRef.current?.expand();
-		} else {
-			collapsableRef.current?.collapse();
-		}
-	}, [isVisible]);
+	const toggleViewed = useCallback(() => {
```
grzpab, line 30: what was wrong here? also, you have no guarantee that the `setVisible` function's callback is going to be run once, it's should not have any side effects

hbjORbj:
```jsx
	useEffect(() => {
		if (isVisible) {
			collapsableRef.current?.collapse();
		} else {
			collapsableRef.current?.expand();
		}	
	}, [isVisible])
```

runs on the initial mount, so even though we initialise `isVisible` to `true`, the views were collapsed. If we want to stick with `useEffect`, we can do

```jsx
	const mountedRef = useRef(false);

	useEffect(() => {
		if (!mountedRef.current) {
			return;
		}
		if (isVisible) {
			collapsableRef.current?.expand();
		} else {
			collapsableRef.current?.collapse();
		}	
	}, [isVisible])

	useEffect(() => {
		mountedRef.current = true;
	}, [])
```

PR #259, opened 2023-04-18
Refactor

hbjORbj:
Follow-up of comments in https://github.com/intuita-inc/intuita-vscode-extension/pull/258


intuita-webview/src/jobDiffView/DiffViewer/index.tsx
```diff
@@ -28,16 +29,24 @@ export const JobDiffView = ({
 	const [isVisible, setVisible] = useState(true);
```
grzpab, line 29: can you switch this again to `false`?

grzpab (approved):

PR #260, opened 2023-04-18
Feature/checkout branch before apply changes

DmytroHryshyn:
- checkout the branch before applying changes to the file system

### DIFF
--- a/src/cases/caseManager.ts
+++ b/src/cases/caseManager.ts
@@ -48,6 +48,10 @@ export class CaseManager {
 		return this.#cases.get(caseHash);
 	}
 
+	public getCaseCount(): number {
+		return this.#cases.size;
+	}
+
 	public getCaseHashJobHashSetValues(): IterableIterator<string> {
 		return this.#caseHashJobHashSetManager.getSetValues();
 	}
--- a/src/cases/caseManager.ts
+++ b/src/cases/caseManager.ts
@@ -44,6 +44,10 @@ export class CaseManager {
 		return this.#cases.values();
 	}
 
+	public getCase(caseHash: CaseHash): Case | undefined {
+		return this.#cases.get(caseHash);
+	}
+
 	public getCaseHashJobHashSetValues(): IterableIterator<string> {
 		return this.#caseHashJobHashSetManager.getSetValues();
 	}
--- a/src/cases/getCaseUniqueName.ts
+++ b/src/cases/getCaseUniqueName.ts
@@ -1,0 +1,5 @@
+import { Case } from "./types"
+
+export const getCaseUniqueName = (kase: Case): string => {
+    return `${kase.hash}_${kase.codemodName}`;
+}
--- a/src/cases/getCaseUniqueName.ts
+++ b/src/cases/getCaseUniqueName.ts
@@ -1,5 +1,5 @@
-import { Case } from "./types"
+import { Case } from './types';
 
 export const getCaseUniqueName = (kase: Case): string => {
-    return `${kase.hash}_${kase.codemodName}`;
-}
+	return `${kase.hash}_${kase.codemodName}`;
+};
--- a/src/components/webview/repository.ts
+++ b/src/components/webview/repository.ts
@@ -1,25 +1,6 @@
 import { API, Repository, Branch, Change, Remote } from '../../types/git';
 import { MessageBus, MessageKind } from '../messageBus';
 
-const branchNameFromStr = (str: string): string => {
-	let branchName = str
-		.toLowerCase()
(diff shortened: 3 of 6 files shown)


src/cases/caseManager.ts
```diff
@@ -44,6 +44,14 @@ export class CaseManager {
 		return this.#cases.values();
 	}
 
+	public getCase(caseHash: CaseHash): Case | undefined {
```
grzpab, line 47: I would say `Case | null` fits here better (`this.#cases` cannot have `null` values for any key either way)

src/components/webview/repository.ts
```diff
@@ -43,6 +44,24 @@ export class RepositoryService {
...
 	}
 
+	public isStackedBranchesEmpty() {
+		return this.__stackedBranches.length === 0;
+	}
+
+	public addStackedBranch(branchName: string): string[] {
+		this.__stackedBranches.push(branchName);
+
+		return this.__stackedBranches;
```
grzpab, line 54: why is this array returned? I think it could return void

src/components/webview/repository.ts
```diff
@@ -76,8 +95,8 @@ export class RepositoryService {
 		);
 	}
 
-	public getBranchName(jobHash: string, jobTitle: string): string {
-		return branchNameFromStr(`${jobTitle}-${jobHash}`);
+	public getBranchName(str: string): string {
+		return branchNameFromStr(str);
```
grzpab, line 99: can you inline `branchNameFromStr` function here?

src/components/webview/repository.ts
```diff
@@ -97,10 +116,25 @@ export class RepositoryService {
...
+		} else {
+			await this.__repo.createBranch(branchName, true);
+		}
+	}
+
 	public async submitChanges(
 		branchName: string,
 		remoteName: string,
 	): Promise<void> {
+		// this should throw instead returning undefined
```
grzpab, line 137: is that a TODO?

src/extension.ts
```diff
@@ -1216,6 +1215,32 @@ export async function activate(context: vscode.ExtensionContext) {
...
+			}
+
+			if (stackedBranchesEmpty) {
+				repositoryService.addStackedBranch(currentBranch.name);
+			}
+
+			/**
+			 * checkout the branch before applying changes to the file system
+			 */
+			const theCase = caseManager.getCase(caseHash as CaseHash);
```
grzpab, line 1233: I usually use `kase` because `case` is a keyword

src/cases/getCaseUniqueName.ts
```diff
@@ -0,0 +1,5 @@
+import { Case } from './types';
+
+export const getCaseUniqueName = (kase: Case): string => {
```
grzpab, line 3: perhaps `buildCaseName` is better?

src/extension.ts
```diff
@@ -463,13 +464,25 @@ export async function activate(context: vscode.ExtensionContext) {
 					throw new Error('Nothing to commit');
 				}
 
-				const treeItem = await treeDataProvider.getTreeItem(
-					jobHash as ElementHash,
-				);
+				const theCase = caseManager.getCase(caseHash as CaseHash);
```
grzpab, line 467: You case use `kase` in that __case__ (pun intended) since `case` is a reserved keyword.
Bonus points for that as well since __Käse__ means cheese in German :D

src/components/webview/repository.ts
```diff
@@ -43,6 +44,24 @@ export class RepositoryService {
 		});
 	}
 
+	public isStackedBranchesEmpty() {
```
grzpab, line 47: `areStackedBranchesEmpty`

src/components/webview/repository.ts
DmytroHryshyn: just left a comment for myself. In  prev version these methods were throwing errors, but then we changed it to return null. I think throwing errors here makes more sense.

src/utilities.ts
```diff
@@ -80,6 +80,25 @@ export const getElementIconBaseName = (kind: Element['kind']): string => {
...
+		.replace(/\s+/g, '-')
+		.replace(/[^a-z0-9-]/g, '-')
+		.replace(/--+/g, '-')
+		.replace(/^-+|-+$/g, '');
+
+	if (branchName.length > 63) {
+		branchName = branchName.substr(0, 63);
+	}
+
+	if (!/^[a-z0-9]/.test(branchName)) {
```
grzpab, line 95: perhaps just `base64url` the branch name then?

grzpab (approved):

PR #266, opened 2023-04-19
imporve PR title and description

DmytroHryshyn:


src/utilities.ts
```diff
@@ -118,3 +118,17 @@ export const buildTreeRootLabel = (caseLabel: string | null) => {
 
 	return `Upgrade ${capitalize(framework)} to v${version}`;
 };
+
+export const buildStackedBranchPRMessage = (
+	stackedBranches: string[],
```
grzpab, line 123: `readonly string[]`

src/utilities.ts
```diff
@@ -118,3 +118,17 @@ export const buildTreeRootLabel = (caseLabel: string | null) => {
...
+): string => {
+	let message = `Current dependencies on/for this PR: \n`;
+
+	message += stackedBranches.reduce((acc, branchName, i) => {
```
grzpab, line 127: can you replace reduce with `forEach`?

grzpab (approved):

PR #267, opened 2023-04-19
feat: integrate monaco with diff-view

henoktsegayetoptal:


src/components/webview/WebviewResolver.ts
```diff
@@ -49,6 +50,27 @@ export class WebviewResolver {
 			`${webviewName}.js`,
 		]);
 
+		// TODO: enable importing chunks on demand
+		// find files that end with .chunk.js
+		const chunkFiles = fs
```
grzpab, line 55: @henoktsegayetoptal @DmytroHryshyn did you discuss the chunkification?

henoktsegayetoptal: I haven't actually , here I am loading chunks at start

DmytroHryshyn (approved):

src/components/webview/WebviewResolver.ts
henoktsegayetoptal: we are generating chunks from monaco loader

intuita-webview/src/shared/Snippet/detectTheme.ts
```diff
@@ -0,0 +1,11 @@
+export function detectBaseTheme(): 'vs-light' | 'vs-dark' {
+	const attribute = document.body.getAttribute('data-vscode-theme-kind');
+	if (
+		attribute &&
```
grzpab, line 4: I think we can then drop `attribute &&` am I right?

henoktsegayetoptal: Yeah that is true

grzpab (approved):

PR #272, opened 2023-04-20
improvement on UI on diff view

henoktse:


DmytroHryshyn (approved):

intuita-webview/src/jobDiffView/DiffViewer/Container.tsx
```diff
@@ -65,7 +67,11 @@ export const Container = forwardRef<HTMLDivElement, ContainerProps>(
 );
 
 type HeaderProps = Readonly<{
+	diffChanges: DiffChanges | null;
```
grzpab, line 70: I would call this structure `DiffChange` or `Diff`, the reason is that plural usually means the structure is an collection.

intuita-webview/src/jobDiffView/DiffViewer/Diff.tsx
```diff
@@ -1,14 +1,58 @@
...
+	const getDiffChanges = (): DiffChanges | undefined => {
+		if (!editorRef.current) {
+			return;
+		}
+		const lineChanges = editorRef.current.getLineChanges();
+		if (!lineChanges) {
+			return;
+		}
+
+		const addedAndDeletedCount = lineChanges.reduce(
```
grzpab, line 24: I would use `forEach` instead of reduce with `added` and `removed` outside of the callback for better readability.

grzpab: Perhaps you could move these calculations to a separate function.

intuita-webview/src/shared/Snippet/DiffEditor.tsx
```diff
@@ -1,13 +1,38 @@
...
+	DiffViewerProps
+>(({ onRefSet, ...props }, ref) => {
 	const theme = useTheme();
+
 	return (
 		<div className="relative w-full h-full">
-			<DiffEditor theme={theme} {...props} />
+			<DiffEditor
+				onMount={(editor) => {
+					typeof ref === 'function' && ref(editor);
```
grzpab, line 19: why's this type check needed here?

intuita-webview/src/jobDiffView/DiffViewer/Container.tsx
```diff
@@ -86,14 +100,41 @@ export const Header = ({
...
+					<h3 className="ml-3 my-0"> {newFileTitle} </h3>
+				</div>
 
 				<div
 					className="flex gap-4"
 					onClick={(e) => {
 						e.stopPropagation();
 					}}
 				>
+					<div className="ml-10 flex items-center">
```
grzpab, line 123: can you display this block if `diff` is not `null`?

intuita-webview/src/shared/Snippet/calculateDiff.tsx
```diff
@@ -0,0 +1,32 @@
+import { monaco } from './DiffEditor';
+
+export type Diff = { added: number; removed: number };
+
+export const getDiff = (lineChanges: monaco.editor.ILineChange[]): Diff => {
+	const diff = new Map<keyof Diff, number>();
```
grzpab, line 6: hmm could you create an `added` and `removed` numbers for readability? Since we have two properties, the `map` seem not to be needed.

grzpab (approved):

PR #273, opened 2023-04-20
wip

DmytroHryshyn:


hbjORbj (changes requested):

intuita-webview/src/sourceControl/CommitView/index.tsx
```diff
@@ -110,51 +123,50 @@ const CreatePR = ({
...
-					<label htmlFor="baseBranch">Base branch:</label>
-					<VSCodeDropdown
-						id="baseBranch"
-						value={baseBranch}
-						onChange={onChangeFormField('baseBranch')}
+				<CommitForm
+					formData={formData}
+					onChangeFormField={onChangeFormField}
+				/>
+				{/* @TODO should we even allow to unapply all jobs? */}
```
grzpab, line 130: We could, but also we might get rid of this feature because Alex said it made little sense to him.

src/extension.ts
```diff
@@ -588,6 +477,25 @@ export async function activate(context: vscode.ExtensionContext) {
...
+
+						const stagedJobHashes = new Set(
+							stagedJobs.map(({ hash }) => hash as JobHash),
+						);
+						messageBus.publish({
+							kind: MessageKind.acceptJobs,
+							jobHashes: stagedJobHashes,
+						});
+
+						// @TODO!!! temp workaround, currently no way to "await" for jobs to be applied to the file system
```
grzpab, line 496: what do you suggest as a solution? I guess we could await a message that confirms that jobs have been accepted?

src/utilities.ts
```diff
@@ -131,3 +131,11 @@ export const buildStackedBranchPRMessage = (
 
 	return message;
 };
+
+export const wait = (ms: number): Promise<void> => {
```
grzpab, line 135: didn't we have this function somewhere already?

DmytroHryshyn: not sure about that, but i will remove this after i remove that hack with waiting for applied jobs

grzpab (approved):

hbjORbj (approved):

PR #274, opened 2023-04-20
UI for git branch / commit integration

hbjORbj:
Screenshot 2023-04-20 at 11 27 29

This PR doesn't include any business logic but UI only.

The vision is the following:
- Each folder is a git branch. Each case inside a folder is a commit.
- User expanding a folder is a signal that he/she wants separate branches for each subfolder.

### DIFF
--- a/intuita-webview/src/main/Tree/index.tsx
+++ b/intuita-webview/src/main/Tree/index.tsx
@@ -1,7 +1,7 @@
 import ReactTreeView from 'react-treeview';
 import { ReactNode, memo, useState } from 'react';
 import { TreeNode } from '../../../../src/components/webview/webviewEvents';
-import { generateRandomColor } from '../../utilities';
+import { buildHash, generateColor } from '../../utilities';
 
 type Props = {
 	index: number;
@@ -30,7 +30,7 @@ type Props = {
 
 const Tree = ({
 	node,
-	depth = 0,
+	depth,
 	renderItem,
 	color: colorProp,
 	index,
@@ -42,7 +42,9 @@ const Tree = ({
 		'acceptedFolderElement',
 	].includes(node.kind);
 	const [open, setIsOpen] = useState(depth === 0);
-	const [color] = useState(hasNoChildren ? colorProp : generateRandomColor());
+	const [color] = useState(
+		hasNoChildren ? colorProp : generateColor(buildHash(node.id)),
+	);
 	const treeItem = renderItem({
 		index,
 		isLastChild,
@@ -63,7 +65,7 @@ const Tree = ({
 	// separate folder children and caseByFolder children
 	// since we want to display all caseByFolder children at the current depth
 	// while we want to display all folder children at 1 level deeper depth
-	if (node.kind && isFolderBreakdown) {
+	if (isFolderBreakdown) {
 		node.children.forEach((element) => {
 			if (!element.kind) {
 				return;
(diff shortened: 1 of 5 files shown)


intuita-webview/src/utilities.ts
```diff
@@ -0,0 +1,2 @@
+export const generateRandomColor = () =>
```
grzpab, line 1: Colors should not be random, because this provides no reproducibility.

One idea is to create a fixed palette (e.g. 32 colors), create a hash of whatever object you want to assign a color to, convert the hash to a number (perhaps the least significant byte of the hash?) and then modulo divide it by 32, getting a color.

intuita-webview/src/main/Tree/index.tsx
```diff
@@ -1,41 +1,69 @@
...
+		isLastChild: boolean;
 	}): ReactNode;
+	color: string;
+	isLastChild: boolean;
 };
 
-const Tree = ({ node, depth = 0, renderItem }: Props) => {
+const Tree = ({
+	node,
+	depth = 0,
```
grzpab, line 33: I would not use default values at all here.

intuita-webview/src/main/Tree/index.tsx
```diff
@@ -1,41 +1,69 @@
...
+	index,
+	isLastChild,
+}: Props) => {
+	const hasNoChildren = !node.children || node.children.length === 0;
+	const isFolderBreakdown = [
+		'folderElement',
+		'acceptedFolderElement',
+	].includes(node.kind);
 	const [open, setIsOpen] = useState(depth === 0);
+	const [color] = useState(hasNoChildren ? colorProp : generateRandomColor());
```
grzpab, line 45: as I mentioned in `utilities`, using randomness makes the experience not reproducible

intuita-webview/src/main/Tree/index.tsx
```diff
@@ -1,41 +1,69 @@
...
-	// separate folder elements and caseByFolder elements
-	// since we want to display all caseByFolder elements above folder elements
-	if (
-		node.kind &&
-		['folderElement', 'acceptedFolderElement'].includes(node.kind)
-	) {
+	// separate folder children and caseByFolder children
+	// since we want to display all caseByFolder children at the current depth
+	// while we want to display all folder children at 1 level deeper depth
+	if (node.kind && isFolderBreakdown) {
```
grzpab, line 66: can `node.kind` be set to `null`?

hbjORbj: Noope, we can remove `node.kind`. Good catch!

intuita-webview/src/utilities.ts
hbjORbj: addressed

intuita-webview/src/main/Tree/index.tsx
hbjORbj: addressed

hbjORbj:
@grzpab 

https://user-images.githubusercontent.com/32841130/233370165-cbb4ac20-f9f4-4051-aaae-d2d7ebf16034.mov

hbjORbj:
final UI

Screenshot 2023-04-20 at 20 08 29

hbjORbj:
Closing this PR because this feature is de-prioritised.

PR #279, opened 2023-04-21
Explore by folder in Codemod-first breakdown

hbjORbj:
### Linear: https://linear.app/intuita/issue/INT-843/explore-by-folder-in-codemod-first-breakdown

> - Allow user to explore by folder in Codemod-first breakdown
> - Consequently, we no longer display the list of individual jobs under each case

> New design is:
> - For each case, there is a root folder with subfolders. 
> - Each folder has "Commit" / "Discard" button.

### Result
https://user-images.githubusercontent.com/32841130/233623062-7e7bb585-e5ce-4a45-b050-d02b6b9b8d96.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -405,12 +405,12 @@
 			"view/title": [
 				{
 					"command": "intuita.caseBreakdown",
-					"when": "view == intuitaMainWebviewExplorer || view == intuitaMainWebview",
+					"when": "false",
 					"group": "navigation"
 				},
 				{
 					"command": "intuita.folderBreakdown",
-					"when": "view == intuitaMainWebviewExplorer || view == intuitaMainWebview",
+					"when": "false",
 					"group": "navigation"
 				},
 				{
--- a/src/components/webview/MainWebviewProvider.ts
+++ b/src/components/webview/MainWebviewProvider.ts
@@ -674,13 +674,13 @@ export class IntuitaProvider implements WebviewViewProvider {
 	private __buildCaseTree = (element: CaseElement): TreeNode => {
 		const actions = [
 			{
-				title: '✗ Discard',
-				command: 'intuita.rejectCase',
+				title: '✓ Commit',
+				command: 'intuita.acceptCase',
 				arguments: [element.hash],
 			},
 			{
-				title: '✓ Commit',
-				command: 'intuita.acceptCase',
+				title: '✗ Discard',
+				command: 'intuita.rejectCase',
 				arguments: [element.hash],
 			},
 		];
--- a/src/components/webview/MainWebviewProvider.ts
+++ b/src/components/webview/MainWebviewProvider.ts
@@ -147,6 +147,8 @@ export class IntuitaProvider implements WebviewViewProvider {
 
 		if (element.kind === ElementKind.CASE) {
 			if (this.__codemodMap.has(element.codemodName)) {
+				// This conditional never should be true.
+				// If it is true, it means there are `CASE` elements with the same codemod name.
 				return;
 			}
 			this.__codemodMap.set(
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -405,12 +405,12 @@
 			"view/title": [
 				{
 					"command": "intuita.caseBreakdown",
-					"when": "view == intuitaMainWebviewExplorer || view == intuitaMainWebview",
+					"when": "false",
```
grzpab, line 408: hang on, then this command will be invisible right?

src/components/webview/MainWebviewProvider.ts
```diff
@@ -118,42 +120,122 @@ export class IntuitaProvider implements WebviewViewProvider {
...
+				children,
 			};
 		}
 
 		if (element.kind === ElementKind.CASE) {
-			mappedNode = {
-				...mappedNode,
-				...this.__buildCaseTree(element),
-			};
+			if (this.__codemodMap.has(element.codemodName)) {
```
grzpab, line 149: this means the elements are up to date?

src/extension.ts
```diff
@@ -1309,20 +1309,133 @@ export async function activate(context: vscode.ExtensionContext) {
...
 			'intuita.acceptFolder',
-			async (arg0, ...otherArgs) => {
-				const firstJobHash: string | null =
-					typeof arg0 === 'string' ? arg0 : null;
-				if (firstJobHash === null) {
-					throw new Error(
-						'Did not pass the jobHashes into the command.',
+			async (_arg0) => {
+				try {
+					const arg0: {
```
grzpab, line 1314: you can actually use a `io-ts` codec for that one

src/extension.ts
```diff
@@ -1309,20 +1309,133 @@ export async function activate(context: vscode.ExtensionContext) {
...
+
+						if (job === null) {
+							continue;
+						}
+
+						stagedJobs.push({
+							hash: job.hash.toString(),
+							label: buildJobElementLabel(
+								job,
+								vscode.workspace.workspaceFolders?.[0]?.uri
```
grzpab, line 1371: this should be calculate before the for-each loop

src/extension.ts
```diff
@@ -1309,20 +1309,133 @@ export async function activate(context: vscode.ExtensionContext) {
...
+							initialData,
+							viewColumn: vscode.ViewColumn.One,
+							webviewName: 'sourceControl',
+						},
+						messageBus,
+					);
+
+					await panelInstance.render();
+
+					const remotes = repositoryService.getRemotes();
```
grzpab, line 1404: `const remotes = repositoryService.getRemotes() ?? [];`
Though `getRemotes` should return an empty array by default.

package.json
hbjORbj: yes, to my understanding, we don't need this command unless we provide options between caseBreakdown vs folderBreakdown.

src/components/webview/MainWebviewProvider.ts
hbjORbj: This conditional never should be true. If it is true, that means we have `CASE` elements that have the same codemod name. Will leave this as a comment.

src/extension.ts
DmytroHryshyn: getRemotes already returns an empty array @grzpab

src/extension.ts
DmytroHryshyn: Can you explain? Here we just iterate jobs and build labels for each job

src/extension.ts
hbjORbj: addressed

DmytroHryshyn (approved):

grzpab (approved):

PR #281, opened 2023-04-21
Feature/add applied unapplied job state

DmytroHryshyn:


src/components/webview/DiffWebviewPanel.ts
```diff
@@ -77,7 +77,10 @@ export class DiffWebviewPanel extends IntuitaWebviewPanel {
 		}
 
 		const job = this.__jobManager.getJob(jobHash);
-		const jobAccepted = this.__jobManager.isJobAccepted(jobHash);
+		// @TODO
```
grzpab, line 80: why's there a TODO?

src/components/webview/MainWebviewProvider.ts
```diff
@@ -533,6 +517,8 @@ export class IntuitaProvider implements WebviewViewProvider {
 	}
 
 	private __onDidReceiveMessage = (message: WebviewResponse) => {
+		console.log(message, 'test');
```
grzpab, line 520: can you remove it?

src/extension.ts
```diff
@@ -89,7 +88,7 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	const jobManager = new JobManager(
 		persistedState?.jobs.map((job) => mapPersistedJobToJob(job)) ?? [],
-		persistedState?.acceptedJobsHashes as JobHash[],
+		persistedState?.appliedJobsHashes as JobHash[],
```
grzpab, line 91: what's the type of `persistedState?.appliedJobsHashes`?

src/components/webview/DiffWebviewPanel.ts
DmytroHryshyn: I left todo because i did not removed all code that was depending on `jobAccepted`

src/extension.ts
DmytroHryshyn: did not noticed that it can be undefined. fixed that.

PR #285, opened 2023-04-21
Fix/henok/int 836 more improvements on diff view

henoktse:


src/components/webview/WebviewPanel.ts
```diff
@@ -60,6 +60,14 @@ export abstract class IntuitaWebviewPanel {
 		this._attachWebviewEventListeners?.();
 	}
 
+	setTitle = (title: string) => {
+		if (!this._panel) {
```
grzpab, line 64: I think we can replace it with `this._panel?.title = title`

grzpab (approved):

PR #286, opened 2023-04-22
migration to Vite

DmytroHryshyn:


intuita-webview/src/shared/Snippet/DiffEditor.tsx
```diff
@@ -1,28 +1,25 @@
...
-						ref.current = editor;
-					}
-					onRefSet();
-				}}
-				{...props}
-			/>
-		</div>
-	);
-});
+const DiffViewer = forwardRef<editor.IStandaloneDiffEditor, DiffViewerProps>(
```
grzpab, line 6: did only formatting change here?

DmytroHryshyn: changed from `monaco.editor.IStandaloneDiffEditor` to `editor.IStandaloneDiffEditor`, i think when we import with `"*" as monaco`, it bundles the whole module.

PR #289, opened 2023-04-25
add getCodemodList method

DmytroHryshyn:


src/components/engineService.ts
```diff
@@ -139,6 +153,47 @@ export class EngineService {
 		this.#noraRustEngineExecutableUri = message.noraRustEngineExecutableUri;
 	}
 
+	public async getCodemodList(): Promise<Readonly<CodemodEntry[]>> {
+		const executableUri = this.#noraNodeEngineExecutableUri;
+
+		if (!executableUri) {
+			throw new EngineNotFoundError('Node engine not found');
```
grzpab, line 160: `The codemod engine node has not been downloaded yet`

src/components/engineService.ts
```diff
@@ -139,6 +153,47 @@ export class EngineService {
...
+		}
+
+		const childProcess = spawn(
```
grzpab, line 163: `spawn` will be closed automatically when the process finishes right?

DmytroHryshyn: Yes, it should

grzpab (approved):

PR #292, opened 2023-04-25
Feature/int 867 add ability to execute a codemod on path

DmytroHryshyn:


src/extension.ts
```diff
@@ -1286,6 +1286,63 @@ export async function activate(context: vscode.ExtensionContext) {
 		),
 	);
 
+	context.subscriptions.push(
+		vscode.commands.registerCommand(
+			'intuita.executeCodemodWithinPath',
```
grzpab, line 1291: is there an existing command we can already remove?

src/extension.ts
```diff
@@ -1286,6 +1286,63 @@ export async function activate(context: vscode.ExtensionContext) {
...
+						);
+					}
+
+					const codemodList = await engineService.getCodemodList();
+
+					const codemodCodemodName =
+						await vscode.window.showQuickPick(
+							codemodList.map(({ name }) => name),
+							{
+								placeHolder:
```
grzpab, line 1308: their API really has `placeHolder` instead of `placeholder`?

DmytroHryshyn: image
Yes, they call it placeHolder 😄

grzpab (approved):

PR #295, opened 2023-04-25
Feature/int 845 job feedback

DmytroHryshyn:


intuita-webview/src/jobDiffView/DiffViewer/DiffItem.tsx
```diff
@@ -20,6 +21,7 @@ export const JobDiffView = ({
 	oldFileTitle,
 	newFileTitle,
 	title,
+	// @TODO why passing postMessage as prop?
 	postMessage,
```
grzpab, line 25: dependency injection is always better than direct global calls

src/components/webview/DiffWebviewPanel.ts
```diff
@@ -61,6 +80,27 @@ export class DiffWebviewPanel extends IntuitaWebviewPanel {
...
+				title: `${job.codemodName} codemod produces invalid output`,
+				body: buildIssueTemplate(job.codemodName),
+				template: 'report-faulty-codemod.md',
+			};
+
+			const query = new URLSearchParams(queryParams).toString();
+
+			commands.executeCommand(
+				'intuita.redirect',
+				`https://github.com/DmytroHryshyn/test_repo/issues/new?${query}`,
```
grzpab, line 101: think needs to be changed into the proper URL

intuita-webview/src/jobDiffView/DiffViewer/DiffItem.tsx
DmytroHryshyn: in react its very unusual to pass utility methods as  props instead of importing them

DmytroHryshyn: and i think we need DI only when we want to be able to switch implementation, for instance, mock dependencies for unit testing. Does it make sense to inject utility functions that are static?

grzpab (approved):

PR #306, opened 2023-04-26
add sticky bar

DmytroHryshyn:
adds sticky bar on the top level 
image


src/extension.ts
```diff
@@ -218,9 +218,17 @@ export async function activate(context: vscode.ExtensionContext) {
 					if (!viewProps) {
 						return;
 					}
+
+					const job = jobManager.getJob(jobHash);
+
+					if (!job) {
+						throw new Error('Unable to get job');
```
grzpab, line 225: `'Unable to find a job with the job hash ${jobHash}`

src/extension.ts
```diff
@@ -309,9 +318,17 @@ export async function activate(context: vscode.ExtensionContext) {
 					if (!viewProps) {
 						return;
 					}
+
+					const job = jobManager.getJob(jobHashes[0]);
```
grzpab, line 322: is `jobHashes[0]` possibly undefined?

intuita-webview/src/jobDiffView/DiffViewer/Header/index.tsx
```diff
@@ -0,0 +1,53 @@
+import { VSCodeButton } from '@vscode/webview-ui-toolkit/react';
+
+import { ReactComponent as UnifiedIcon } from '../../../assets/Unified.svg';
+import { ReactComponent as SplitIcon } from '../../../assets/Split.svg';
+import { DiffViewType } from '../../../shared/types';
+
+import styles from './style.module.css';
+
+type Props = {
```
grzpab, line 9: `Readonly<`

grzpab:
@DmytroHryshyn can you write in the header something like: `Executed the codemod: ${name or path}`
Also, the button on the right could use the phrase Inline / Side by Side inlined before the image.

DmytroHryshyn:
> @DmytroHryshyn can you write in the header something like: `Executed the codemod: ${name or path}` Also, the button on the right could use the phrase Inline / Side by Side inlined before the image.

I've removed button titles intentionally, wanted to follow vscode design code (they just use icon+hover message)
image

grzpab (approved):

PR #315, opened 2023-04-27
Support Running Multiple Codemods and File Explorer changes depending on which case in selected

hbjORbj:

### DIFF
--- a/src/components/webview/FileExplorerProvider.ts
+++ b/src/components/webview/FileExplorerProvider.ts
@@ -24,7 +24,7 @@ import {
 import { Job, JobHash, JobKind } from '../../jobs/types';
 import { getElementIconBaseName } from '../../utilities';
 import { JobManager } from '../jobManager';
-import { CaseWithJobHashes } from '../../cases/types';
+import { CaseHash, CaseWithJobHashes } from '../../cases/types';
 import {
 	buildJobElement,
 	compareJobElements,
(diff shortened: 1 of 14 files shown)


src/components/webview/FileExplorerProvider.ts
```diff
@@ -100,6 +94,48 @@ export class FileExplorerProvider implements WebviewViewProvider {
 		});
 	}
 
+	public updateExplorerView(caseHash: string | null) {
```
grzpab, line 97: `caseHash: CaseHash`

`null` should be not accepted

src/components/webview/webviewEvents.ts
```diff
@@ -199,6 +199,10 @@ export type WebviewResponse =
 			oldFileContent: string;
 			newFileContent: string;
 	  }>
+	| Readonly<{
+			kind: 'webview.campaignManager.selectCase';
+			hash: string | null;
```
grzpab, line 204: can you unselect a case? if so, the `hash` should be just `CaseHash`

hbjORbj: Currently, the only way to unselect a case is to select another case. I think it's not worth supporting unselecting a case when there is only one case in the panel.

hbjORbj: Regardless, I refactored so that we have `hash: CaseHash;`

src/components/webview/FileExplorerProvider.ts
hbjORbj: addressed

DmytroHryshyn (approved):

PR #321, opened 2023-04-27
remove jobs from diff view when accepted, dispose diff view when no j…

DmytroHryshyn:
https://app.claap.io/intuita/c-BT2DRbCjzO-xyN0kdZiwMR7

### DIFF
--- a/intuita-webview/src/sourceControl/CommitView/index.tsx
+++ b/intuita-webview/src/sourceControl/CommitView/index.tsx
@@ -158,14 +158,16 @@ const CreatePR = ({ loading, initialFormData, remoteOptions }: Props) => {
 					>
 						Commit
 					</VSCodeButton>
-					<VSCodeButton
-						appearance="primary"
-						disabled={loading}
-						className={styles.actionButton}
-						onClick={handleSubmit}
-					>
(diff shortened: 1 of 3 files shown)


intuita-webview/src/sourceControl/CommitView/index.tsx
```diff
@@ -158,14 +158,16 @@ const CreatePR = ({ loading, initialFormData, remoteOptions }: Props) => {
...
 					</VSCodeButton>
-					<VSCodeButton
-						appearance="primary"
-						disabled={loading}
-						className={styles.actionButton}
-						onClick={handleSubmit}
-					>
-						Commit & Pull Request
-					</VSCodeButton>
+					{createNewBranch ? (
```
hbjORbj, line 161: Creating PR is only possible when creating a new branch? Why?

DmytroHryshyn: we always create pr using the current branch as a base branch (because we ran codemod on it), and in case if we committed changes on current branch, base and target branches for pr will be the same which does not make any sense. So to be able to create pr we always need to create a new branch.

PR #335, opened 2023-04-30
Feature/int 931 apply changes from diff view

DmytroHryshyn:
image

### DIFF
--- a/intuita-webview/src/jobDiffView/DiffViewer/Header/index.tsx
+++ b/intuita-webview/src/jobDiffView/DiffViewer/Header/index.tsx
@@ -11,10 +11,12 @@ type Props = Readonly<{
 	title: string;
 	viewType: DiffViewType;
 	jobs: JobDiffViewProps[];
+	diffId: string;
+	changesAccepted: boolean;
 	onViewChange(value: DiffViewType): void;
 }>;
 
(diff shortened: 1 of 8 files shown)


hbjORbj:
Can you rebase on `main`?

intuita-webview/src/jobDiffView/DiffViewer/Header/index.tsx
```diff
@@ -48,19 +59,22 @@ const Header = ({ title, viewType, jobs, onViewChange }: Props) => {
...
+						onClick={handleCommit}
+					>
+						Commit...
+					</VSCodeButton>
+				) : (
+					<VSCodeButton
+						appearance="primary"
+						onClick={handleApplySelected}
+					>
+						Apply selected
```
hbjORbj, line 75: I like `Apply all`.. but this is debatable..

DmytroHryshyn: `Apply selected` looks better to me, because we will be able to select specific jobs to be applied

hbjORbj: Ahh, in that case, yes, agree

grzpab (approved):

PR #343, opened 2023-05-01
make accepted case readonly

DmytroHryshyn:


intuita-webview/src/jobDiffView/DiffViewer/Container.tsx
```diff
@@ -137,13 +144,24 @@ export const Header = ({
...
 					<div
-						className="flex ml-10 justify-between checkbox-container items-center"
+						className={cn(
+							'viewed-button flex ml-10 justify-between checkbox-container items-center',
+							{
+								disabled: changesAccepted,
+							},
+						)}
 						onClick={(e) => {
+							if (changesAccepted) {
```
grzpab, line 154: should we stop event propagation if the changes have been accepted?

DmytroHryshyn: oh, i can remove that check, `.disabled` class already prevents all mouse interactions with that element

grzpab (approved):

PR #344, opened 2023-05-01
Feature/int 938 create a unique id per case stemming

DmytroHryshyn:
image
adds ability to run same codemod multiple times. each case hash is unique now


grzpab (commented):
@DmytroHryshyn do you think we should add `executionId` to the job directly, or doesn't it make any sense?

grzpab (approved):

PR #350, opened 2023-05-02
Feat/int 831 add progress bar to codemod discovery

henoktse:


intuita-webview/src/codemodList/TreeView/index.tsx
```diff
@@ -44,6 +49,25 @@ const TreeView = ({
 	const [focusedNodeId, setFocusedNodeId] = useState('');
 	const [editExecutionPath, setEditExecutionPath] =
 		useState<CodemodTreeNode<string> | null>(null);
+	const [progress, { progressBar, stopProgress }] = useProgressBar();
+	const [executionStack, setExecutionStack] = useState<Set<CodemodHash>>(
```
grzpab, line 53: why is it a set? because at one point, we only can run one codemod?

intuita-webview/src/codemodList/TreeView/index.tsx
```diff
@@ -44,6 +49,25 @@ const TreeView = ({
...
+		new Set(),
+	);
+
+	useEffect(() => {
+		if (progress || executionStack.size === 0) {
+			return;
+		}
+		const hash = Array.from(executionStack).shift();
```
grzpab, line 61: if we use set, we should use the alphanumerical sorting to have this line always yield the same results

intuita-webview/src/codemodList/useProgressBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { CodemodHash, WebviewMessage } from '../shared/types';
+import styles from './TreeView/style.module.css';
+import { vscode } from '../shared/utilities/vscode';
+
+type ProgressType = {
+	progress: number;
+	codemodHash: CodemodHash;
+};
+
+export const useProgressBar = (): [
```
grzpab, line 12: do we know the `codemodHash` when we use the `useProgressBar` hook?

grzpab (approved):

PR #357, opened 2023-05-02
Improvements in File Explorer Panel

hbjORbj:
## Changes
1. Collapse/Expand on pressing the Chevron only, not the item
2. Press on the folder element leads to the first file inside the folder


intuita-webview/src/codemodList/components/Container.tsx
```diff
@@ -36,6 +36,7 @@ export const Container = ({
 			})}
 		>
 			<Collapsable
+				id="codemodListcollapsable"
```
grzpab, line 39: camel-case typo

PR #359, opened 2023-05-03
terminate process

DmytroHryshyn:


src/components/engineService.ts
```diff
@@ -440,7 +443,17 @@ export class EngineService {
...
 
+		let timer: NodeJS.Timeout | null = null;
+
 		interfase.on('line', async (line) => {
+			if (timer !== null) {
+				clearTimeout(timer);
+			}
+
+			timer = setTimeout(() => {
+				childProcess.kill();
```
grzpab, line 454: what happens when we attempt to kill an already terminated process?

DmytroHryshyn: just tried to call `kill` twice
```
childProcess.kill();
childProcess.kill();
```
second one was just ignored

grzpab (approved):

PR #392, opened 2023-05-11
Fix the spinning wheel for loading public codemods when no workspace is open

hbjORbj:
## Linear: https://linear.app/intuita/issue/INT-925/fix-the-spinning-wheel-for-loading-public-codemods-when-no-workspace

### DIFF
--- a/src/packageJsonAnalyzer/codemodService.ts
+++ b/src/packageJsonAnalyzer/codemodService.ts
@@ -16,12 +16,12 @@ import {
 import { EngineService } from '../components/engineService';
 
 export class CodemodService {
-	#rootPath: string | null;
+	#rootPath: string | 'NO_ACTIVE_WORKSPACE';
 	#codemodItemsMap: Map<CodemodHash, CodemodElement> = new Map();
 	#publicCodemods: Map<CodemodHash, CodemodElement> = new Map();
 
 	constructor(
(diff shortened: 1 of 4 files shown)


src/packageJsonAnalyzer/codemodService.ts
```diff
@@ -16,12 +16,12 @@ import {
 import { EngineService } from '../components/engineService';
 
 export class CodemodService {
-	#rootPath: string | null;
+	#rootPath: string | 'NO_ACTIVE_WORKSPACE';
```
grzpab, line 19: I would still keep it nullable

hbjORbj: addressed

src/packageJsonAnalyzer/codemodService.ts
```diff
@@ -16,12 +16,12 @@ import {
 import { EngineService } from '../components/engineService';
 
 export class CodemodService {
-	#rootPath: string | null;
+	#rootPath: string | 'NO_ACTIVE_WORKSPACE' | null;
```
grzpab, line 19: why do we need a difference between a `NO_ACTIVE_WORKSPACE` and null?

grzpab (approved):

PR #395, opened 2023-05-11
track errors and collect events

DmytroHryshyn:
- tracks command execution failure
- remove unused commands
- adds user metrics tracking
image


src/components/jobManager.ts
```diff
@@ -168,9 +168,15 @@ export class JobManager {
 					continue;
 				}
 
+				const deletedJobs = new Set(
```
grzpab, line 171: why is this check needed?

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,35 @@
+export type ErrorEvent = Readonly<{
+	kind: 'failedToExecuteCommand';
+	commandName: string;
+	errorMessage: string;
```
grzpab, line 4: can error message contain file paths or other sensitive information?

DmytroHryshyn: we do not include stack so its more or less safe. but i think we cannot guarantee that some thirdparty lib does not include file name in the message. Do you propose to exclude errorMessage from tracking? Or add some kind of message validation?

src/components/jobManager.ts
DmytroHryshyn: i wanted to pass jobs instead of hashes so i can group jobs by executionId in the tracking service.

src/extension.ts
```diff
@@ -154,6 +156,12 @@ export async function activate(context: vscode.ExtensionContext) {
 		),
 	);
 
+	const telemetryKey = '63abdc2f-f7d2-4777-a320-c0e596a6f114';
```
grzpab, line 159: should this key be publicly available?

src/telemetry/vscodeTelemetry.ts
```diff
@@ -0,0 +1,151 @@
...
+			properties,
+			measurements,
+		};
+	}
+
+	sendEvent(event: Event): void {
+		const { name, properties, measurements } =
+			this.__rawEventToTelemetryEvent(event);
+
+		console.log('sendEvent', name, properties, measurements);
```
grzpab, line 133: can you remove this line?

package.json
```diff
@@ -209,30 +209,14 @@
 					"command": "intuita.clearState",
 					"when": "false"
 				},
-				{
```
grzpab, line 212: should these commands be removed?

src/extension.ts
DmytroHryshyn: yes, its public key

package.json
DmytroHryshyn: yes, this is old code, we cannot accept/reject specific jobs or folders now

grzpab (approved):

PR #406, opened 2023-05-15
Display 1 diff at a time + Keyboard Accessibility Improvements

hbjORbj:
### Linear: https://linear.app/intuita/issue/INT-1023/regression-selecting-filefolder-doesnt-lead-to-the-relevant-diff-view

### DIFF
--- a/intuita-webview/src/jobDiffView/App.tsx
+++ b/intuita-webview/src/jobDiffView/App.tsx
@@ -11,18 +11,25 @@ import { JobDiffViewContainer } from './DiffViewer/index';
 import './index.css';
 import LoadingProgress from './Components/LoadingProgress';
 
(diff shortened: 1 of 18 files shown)


intuita-webview/src/jobDiffView/App.tsx
```diff
@@ -104,9 +115,8 @@ function App() {
 	return (
 		<main className="App">
 			<JobDiffViewContainer
-				scrollIntoHash={scrollIntoHash}
 				diffId={diffId}
-				jobs={data}
+				jobs={[data[jobIndex]!]}
```
grzpab, line 119: you can check explicitly to see if it exists to satisfy the compiler (if it doesn't, provide an empty array)

grzpab (approved):

PR #419, opened 2023-05-19
New Diff View & Change Explorer

hbjORbj:
### Changes
- User can move between Diff view and Folder/File item in Change Explorer panel using keyboard arrow (left and right)
- Apply Selected / Discard buttons & Select/Unselect all checkbox are removed from Diff view
- Apply Selected / Discard buttons & Select/Unselect all checkbox are added to Change Explorer
- Each file item in Change explorer now has a checkbox (for applying/unapplying file change)
- Each file item in Change explorer has a different background color to differentiate it from folder items

### Claap: https://app.claap.io/intuita/new-diff-view-and-change-explorer-c-BT2DRbCjzO-P7J4e5EFxVZo

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1027,6 +1027,7 @@ export async function activate(context: vscode.ExtensionContext) {
 				if (caseHash === null) {
 					return;
 				}
+				fileExplorerProvider.setCaseHash(caseHash);
 				fileExplorerProvider.showView();
 				fileExplorerProvider.updateExplorerView(caseHash);
 			},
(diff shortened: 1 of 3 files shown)


src/components/webview/DiffWebviewPanel.ts
```diff
@@ -117,26 +117,6 @@ export class DiffWebviewPanel extends IntuitaWebviewPanel {
 			);
 		}
 
-		if (message.kind === 'webview.global.applySelected') {
```
grzpab, line 120: why are these handlers removed?

src/extension.ts
```diff
@@ -343,6 +343,36 @@ export async function activate(context: vscode.ExtensionContext) {
 		),
 	);
 
+	context.subscriptions.push(
+		vscode.commands.registerCommand(
+			'intuita.disposeView',
```
grzpab, line 348: what does it do?

grzpab (approved):

PR #427, opened 2023-05-22
Codemod Discovery Edit Path Flow changes

hbjORbj:
### Changes
- each codemod has its own execution path
- we show the path after the codemod name when hovering the on menu item as "(pencil icon) Path: '''''
- by default, the path is ./ instead of /
- We can modify the path by clicking on the path input and replacing it and confirming with "Enter" key. We can cancel modifying with `Escape` key.
- If user tries to modify path to an invalid path, we show an error popup and the path stays the same.

### DIFF
--- a/intuita-webview/src/codemodList/components/DirectorySelector.tsx
+++ b/intuita-webview/src/codemodList/components/DirectorySelector.tsx
@@ -54,7 +54,7 @@ export const DirectorySelector = ({
 		}
 
 		if (event.key === 'Enter') {
-			if (value.length <= 2) {
+			if (value.length < 2) {
 				// "./" (default path) should always be there
 				handleCancel();
 				return;


intuita-webview/src/codemodList/TreeView/TreeItem.tsx
```diff
@@ -19,23 +24,90 @@ type Props = {
...
 	open,
 	focused,
+	rootPath,
 	actionButtons,
 	hasChildren,
 	onClick,
 	depth,
+	executionPath,
 }: Props) => {
+	const [targetPath, setTargetPath] = useState('');
```
grzpab, line 47: does this come from (or should come from) the workspace state?
Edit: I see this is a volatile path (you could even call it volatile if you want)
Edit2: but where is it set?

src/components/webview/CodemodListProvider.ts
```diff
@@ -296,6 +306,9 @@ export class CodemodListPanelProvider implements WebviewViewProvider {
 					},
 				],
 				children: [],
+				...(executionPath !== null && {
```
grzpab, line 309: it cannot be nullable anymore

src/persistedState/codecs.ts
```diff
@@ -43,3 +43,25 @@ export const persistedStateCodec = buildTypeCodec({
...
 
 export type PersistedState = t.TypeOf<typeof persistedStateCodec>;
+
+export const syntheticErrorCodec = buildTypeCodec({
+	kind: t.literal('syntheticError'),
+	message: t.string,
+});
+
+export const workspaceStateCodec = t.union([
+	buildTypeCodec({
```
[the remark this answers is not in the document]
grzpab: I wanted to have 1:1 pairing between the structures of fp-ts and the saved data, because the former are serializable. This way we need no casting.

grzpab (approved):

PR #451, opened 2023-05-30
Add keyboard accessibility to discovery panel

hbjORbj:
- implements arrow key (up, down) to move between codemods
- Linear: https://linear.app/intuita/issue/INT-1106/add-keyboard-accessibility-in-codemod-discovery
- Claap: https://app.claap.io/intuita/add-keyboard-accessibility-to-discovery-panel-c-BT2DRbCjzO-Jq4sL5Go1E5T

### DIFF
--- a/src/components/webview/CodemodListProvider.ts
+++ b/src/components/webview/CodemodListProvider.ts
@@ -421,6 +421,8 @@ export class CodemodListPanelProvider implements WebviewViewProvider {
 
 		this.__treeMap.set(hash, node);
 
+		// children is set after adding the node to the tree map
+		// in order to retain the ordering
 		node.children = children.map((child) => this.__getTreeNode(child));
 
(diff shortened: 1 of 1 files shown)


src/components/webview/CodemodListProvider.ts
```diff
@@ -396,20 +402,28 @@ export class CodemodListPanelProvider implements WebviewViewProvider {
...
 			id: hash,
 			uri: path,
 			actions: [],
-			children: children.map((child) => this.__getTreeNode(child)),
+			children: [],
 		};
+
+		this.__treeMap.set(hash, node);
+
+		node.children = children.map((child) => this.__getTreeNode(child));
```
grzpab, line 424: why are the children calculated after setting the node in the tree map?

grzpab (approved):

PR #455, opened 2023-05-31
INT-1121 persist the public codemods expansion into workspace state

grzpab:
https://app.claap.io/intuita/persist-the-public-codemods-expansion-state-c-BT2DRbCjzO-GpuXMf_-cDbL


intuita-webview/src/codemodList/App.tsx
```diff
@@ -45,12 +44,27 @@ function App() {
 		};
 	}, []);
 
+	const setPublicCodemodsExpanded = useCallback(
```
DmytroHryshyn, line 47: lets move pure functions outside of component scope instead of wrapping in useCallback

PR #467, opened 2023-06-02
Viewed toggle getting unset when going back and forth between file diffs

hbjORbj:
### Problem

Currently, we always reset the state of diff view to the default value when rendering it. This is a bug originating from this. This PR makes it to store the diff view state in an object properly with job hash as a key. 

### Claap: https://app.claap.io/intuita/fix-for-viewed-toggle-getting-unset-when-going-back-and-forth-between-file-diffs-c-BT2DRbCjzO-FTVIsLwwCg6h

### DIFF
--- a/intuita-webview/src/jobDiffView/DiffViewer/index.tsx
+++ b/intuita-webview/src/jobDiffView/DiffViewer/index.tsx
@@ -49,6 +49,14 @@ type DiffData = Record<JobHash, DiffItem>;
 
 const defaultHeight = 1200;
 
+const jobDiffViewDefaultState = {
+	visible: true,
+	diff: null,
+	height: defaultHeight,
+	expanded: true,
+	containerHeight: 50,
+};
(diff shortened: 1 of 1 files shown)


intuita-webview/src/jobDiffView/DiffViewer/index.tsx
```diff
@@ -85,18 +86,16 @@ export const JobDiffViewContainer = ({
...
-			};
-			return acc;
-		}, {} as DiffData);
-		prevDiffData.current = diffData;
-		setDiffData(diffData);
+		setDiffData((prevDiffData) => {
+			const data = { ...prevDiffData };
+			jobs.forEach((job) => {
+				const hash = job.jobHash;
+				if (!data[hash]) {
```
grzpab, line 93: can you explain this check?

grzpab (approved):

PR #474, opened 2023-06-05
Feature/int 1158 ux job diff view scroll position

DmytroHryshyn:
- removes code related to virtualization and height measurement
 - scrolls editor to the top when content is changed
 Claap: 
 https://app.claap.io/intuita/c-BT2DRbCjzO-kS-urJOnFmba


grzpab (approved):
Tested locally.

PR #483, opened 2023-06-06
Further keyboard a11y improvements in discovery

hbjORbj:
#### Problem

when a folder is collapsed and you hit down arrow the selection wont be in the view and instead it seems that i goes into the children of that folder instead.

### DIFF
--- a/intuita-webview/src/codemodList/TreeView/Tree.tsx
+++ b/intuita-webview/src/codemodList/TreeView/Tree.tsx
@@ -7,7 +7,7 @@ const getIndex = (
 	nodes: ReadonlyArray<CodemodTreeNode>,
 	hash: CodemodHash | null,
 ): number => {
-	return nodes.findIndex((_node) => hash === _node.id) ?? null;
+	return nodes.findIndex((_node) => hash === _node.id);
 };
 
 type Props = {
--- a/intuita-webview/src/codemodList/TreeView/Tree.tsx
+++ b/intuita-webview/src/codemodList/TreeView/Tree.tsx
@@ -128,16 +128,17 @@ const Tree = ({
 			const nodesAtNextDepth: ReadonlyArray<CodemodTreeNode> =
 				nodesByDepth[depth + 1] ?? [];
 			const lastChildInNodesByDepth =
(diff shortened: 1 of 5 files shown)


intuita-webview/src/codemodList/TreeView/Tree.tsx
```diff
@@ -3,6 +3,19 @@ import { ReactNode } from 'react';
...
 
+const getIndex = (
+	nodes: ReadonlyArray<CodemodTreeNode>,
+	hash: CodemodHash | null,
+): number | null => {
+	if (hash === null) {
+		return null;
+	}
+
+	const indexInNodesByDepth =
```
grzpab, line 14: 1) return immediately in this line
2) findIndex doesn't yield `null`, but `-1` in case of a problem

intuita-webview/src/codemodList/TreeView/Tree.tsx
```diff
@@ -40,8 +55,101 @@ const Tree = ({
...
+		if (node.id !== focusedId) {
+			return;
+		}
+
+		const nodesAtCurrentDepth = nodesByDepth[depth] ?? [];
+		const indexAmongNodesAtCurrentDepth = getIndex(
+			nodesAtCurrentDepth,
+			node.id,
+		);
+		if (indexAmongNodesAtCurrentDepth === null) {
```
grzpab, line 67: `=== -1`

intuita-webview/src/codemodList/TreeView/Tree.tsx
```diff
@@ -3,6 +3,13 @@ import { ReactNode } from 'react';
 import { CodemodHash, CodemodTreeNode } from '../../shared/types';
 import { useKey } from '../../jobDiffView/hooks/useKey';
 
+const getIndex = (
+	nodes: ReadonlyArray<CodemodTreeNode>,
+	hash: CodemodHash | null,
+): number => {
+	return nodes.findIndex((_node) => hash === _node.id) ?? null;
```
grzpab, line 10: doesn't return `null`, just `-1`

grzpab (approved):

PR #495, opened 2023-06-07
Refactor FileExplorer

hbjORbj:
- refactors `FileExplorerProvider`, removing unnecessary code
- initialises `viewProps` of File explorer panel to `window.INITIAL_STATE.viewProps`

### DIFF
--- a/src/components/webview/FileExplorerProvider.ts
+++ b/src/components/webview/FileExplorerProvider.ts
@@ -44,6 +44,8 @@ import { CaseManager } from '../../cases/caseManager';
 import { DiffWebviewPanel } from './DiffWebviewPanel';
 import { WorkspaceState } from '../../persistedState/workspaceState';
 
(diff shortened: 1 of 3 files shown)


src/components/webview/FileExplorerProvider.ts
```diff
@@ -74,47 +75,67 @@ export class FileExplorerProvider implements WebviewViewProvider {
...
+			);
 		});
 
 		this.__attachExtensionEventListeners();
 		this.__attachWebviewEventListeners();
 	}
 
-	public setView(data: Extract<View, { viewId: 'fileExplorer' }>) {
-		this.__lastData = data;
+	public setView(caseHash: CaseHash | null): void {
```
grzpab, line 108: In the future, `setView` should accept no parameters (because the only goal is to send information to the view). For not it's fine.

PR #496, opened 2023-06-08
fix incorrect state

DmytroHryshyn:


src/components/webview/CampaignManagerProvider.ts
```diff
@@ -273,6 +283,12 @@ export class CampaignManagerProvider implements WebviewViewProvider {
 				this.setView();
 			}
 		});
+
+		this.__webviewView.onDidChangeVisibility(() => {
+			if (this.__webviewView?.visible) {
```
grzpab, line 288: should we apply this check for instance for the Error Provider as well?

grzpab (approved):

PR #497, opened 2023-06-08
persist codemods

DmytroHryshyn:


src/codemods/types.ts
```diff
@@ -0,0 +1,19 @@
...
+	description: t.string,
+	engine: t.union([
+		t.literal('jscodeshift'),
+		t.literal('ts-morph'),
+		t.literal('repomod-engine'),
+		t.literal('filemod-engine'),
+	]),
+});
+
+export const codemodListCodec = t.readonlyArray(codemodEntryCodec);
```
grzpab, line 19: I think this statement is unnecessary, let's keep the number of codeds as small as possible.

PR #499, opened 2023-06-09
Dmytro/int 1157 allow the right side of the diff view to be editable

DmytroHryshyn:
https://app.claap.io/intuita/c-BT2DRbCjzO-_Kfwybvsjo36

### DIFF
--- a/src/components/jobManager.ts
+++ b/src/components/jobManager.ts
@@ -236,9 +236,14 @@ export class JobManager {
 		this.#appliedJobsHashes.delete(jobHash);
 	}
 
-	public async modifyJobContent(jobHash: JobHash, newJobContent: string) {
+	public async changeJobContent(jobHash: JobHash, newJobContent: string) {
 		const job = this.getJob(jobHash);
-		const newContentUri = job?.newContentUri ?? null;
+
+		if (job === null) {
+			return;
+		}
+
+		const newContentUri = job.newContentUri;
 
 		if (newContentUri === null) {
 			return;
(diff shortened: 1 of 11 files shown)


src/jobs/types.ts
```diff
@@ -22,4 +22,5 @@ export type Job = Readonly<{
 	codemodName: string;
 	createdAt: number;
 	executionId: string;
+	modifiedByUser?: boolean;
```
grzpab, line 25: by default that should be false I think, our serializers should make it falsy

src/components/jobManager.ts
```diff
@@ -236,6 +236,21 @@ export class JobManager {
 		this.#appliedJobsHashes.delete(jobHash);
 	}
 
+	public async modifyJobContent(jobHash: JobHash, newJobContent: string) {
```
grzpab, line 239: maybe let's call it `changeJobContent`?

src/components/jobManager.ts
```diff
@@ -236,6 +236,21 @@ export class JobManager {
...
+
+		if (newContentUri === null) {
+			return;
+		}
+
+		await this.__fileService.updateFileContent({
+			uri: newContentUri,
+			content: newJobContent,
+		});
+		this.#jobMap.set(jobHash, { ...job, modifiedByUser: true } as Job);
```
grzpab, line 251: why `as Job` is needed?

src/components/messageBus.ts
```diff
@@ -55,6 +55,8 @@ export const enum MessageKind {
 	focusCodemod = 35,
 
 	afterDryRunHooksExecuted = 36,
+
+	modifyJobContent = 37,
```
grzpab, line 59: perhaps `changedJobContent`?

grzpab:
@DmytroHryshyn I think we should persist only a list of jobs modified by users in the persisted state, this way it's easier to manage job without recreating them.

PR #503, opened 2023-06-09
Feature/int 1055 formatting on view formatting on apply

DmytroHryshyn:


package.json
```diff
@@ -133,6 +133,11 @@
 					"order": 7,
 					"type": "string",
 					"description": "Specify a command that will be executed when dry run is completed"
+				},
+				"intuita.useFormatting": {
```
grzpab, line 137: `intuita.formatWithPrettier`, because "useFormatting" is a command, not a statement

package.json
```diff
@@ -133,6 +133,11 @@
...
+					"order": 8,
+					"type": "boolean",
+					"description": "Use Prettier formatter localized based on the file in question; if the config cannot be found, use the Intuita's Prettier config"
```
grzpab, line 140: Use the Prettier formatter with the configuration derived from the location of a file to be formatter. If none is found, the extension will use the custom Intuita's config.

src/formatter.ts
```diff
@@ -0,0 +1,48 @@
+import { format, resolveConfig, Options } from 'prettier';
```
grzpab, line 1: is prettier installed in package.json as a dependency?

grzpab (approved):

PR #507, opened 2023-06-12
add redux, redux-persist boilerplate

DmytroHryshyn:
- adds redux state
- stores data in the redux state


src/data/storage.ts
```diff
@@ -0,0 +1,39 @@
...
+			const storedValue = this.__memento.get(key);
+			resolve(storedValue);
+		});
+	}
+
+	public setItem(key: string, value: string): Promise<void> {
+		return new Promise((resolve) => {
+			this.__memento.update(key, value);
+
+			resolve(void 0);
```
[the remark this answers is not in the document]
DmytroHryshyn: i think i need to `resolve()` with no value, because the interface of `WebStorage` has return type Promise<void>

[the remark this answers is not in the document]
grzpab: Yes, you are right

src/data/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import MementoStorage from './storage';
+
+import rootReducer, { actions } from './slice';
+import { Memento } from 'vscode';
+
+const buildStore = (workspaceState: Memento) => {
+	const persistConfig = {
+		key: 'root',
+		storage: new MementoStorage(workspaceState),
+		// debugger: console.log,
```
grzpab, line 13: perhaps temporarily we should use `debugger: console.debug` here?

src/data/index.ts
```diff
@@ -0,0 +1,49 @@
...
+	};
+
+	const persistedReducer = persistReducer(persistConfig, rootReducer);
+
+	// @TODO ensure that building with NODE_ENV=production var, otherwise default dev middleware will be included
```
grzpab, line 18: did you figure out how that one should work with our current build system?

src/data/slice.ts
```diff
@@ -0,0 +1,220 @@
...
+	visible: boolean;
+	focusedFileExplorerNodeId: string | null;
+	openedFileExplorerNodeIds: ReadonlyArray<string> | null;
+}>;
+
+type CommunityState = Readonly<{
+	visible: boolean;
+}>;
+
+type State = {
```
grzpab, line 35: I recall you wanted this to become normalized?

src/data/slice.ts
```diff
@@ -0,0 +1,220 @@
...
+	codemodDiscovery: CodemodDiscoveryState;
+	codemodRuns: CodemodRunsState;
+	changeExplorer: ChangeExplorerState;
+	community: CommunityState;
```
grzpab, line 39: should we call it for instance `communityPane`? does it make sense?

src/data/index.ts
DmytroHryshyn: its better to remove this debugger at all, it spams too much

src/data/index.ts
DmytroHryshyn: okay, i need to fix package.json for that

src/data/slice.ts
DmytroHryshyn: normalization is handled by adapters, i use it for entity collection only

src/data/slice.ts
DmytroHryshyn: agree, lets distinguish between domain state and ui state. i will call it communityView and same for other views

src/data/slice.ts
```diff
@@ -0,0 +1,221 @@
...
+
+type CodemodRunsState = Readonly<{
+	selectedCaseHash: string | null;
+	visible: boolean;
+}>;
+
+type ChangeExplorerState = Readonly<{
+	visible: boolean;
+	focusedFileExplorerNodeId: string | null;
+	openedFileExplorerNodeIds: ReadonlyArray<string> | null;
```
grzpab, line 28: this can be not-nullable

grzpab (commented):
I saw that is reports the following issue in the debugger:

```
A non-serializable value was detected in an action, in the path: `register`. Value: ƒ register(key) {
    _pStore.dispatch({
      type: _constants__WEBPACK_IMPORTED_MODULE_0__.REGISTER,
      key: key
    });
  } 
Take a look at the logic that dispatched this action:  {type: 'persist/PERSIST', register: ƒ, rehydrate: ƒ} 
(See https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants) 
(To allow non-serializable values see: https://redux-toolkit.js.org/usage/usage-guide#working-with-non-serializable-data)
```

grzpab (approved):

PR #510, opened 2023-06-15
INT-1190 have the Error Panel use Redux

grzpab:


src/components/webview/ErrorWebviewProvider.ts
```diff
@@ -46,23 +44,43 @@ export class ErrorWebviewProvider implements WebviewViewProvider {
 		);
 
 		messageBus.subscribe(MessageKind.clearState, () => {
-			__workspaceState.setSelectedCaseHash(null);
```
DmytroHryshyn, line 49: other components might use this value from the `workspaceState`. Lets keep it for now

DmytroHryshyn (approved):

PR #517, opened 2023-06-15
[Change Explorer] Debounce rendering diff view based on tree item focus

hbjORbj:
#### Linear: https://linear.app/intuita/issue/INT-1191/[change-explorer]-debounce-rendering-diff-view-based-on-tree-item


intuita-webview/src/fileExplorer/TreeView/index.tsx
```diff
@@ -193,6 +201,16 @@ const TreeView = ({
 		},
 		initializer,
 	);
+	const debouncedOnFileOrFolderSelectedRef = useRef(
```
grzpab, line 204: why do we use a ref here?

grzpab: @hbjORbj can you use `useCallback` instead?

PR #528, opened 2023-06-16
Feature/add state validation

DmytroHryshyn:


src/data/index.ts
```diff
@@ -15,8 +17,23 @@ const buildStore = (workspaceState: Memento) => {
 		rootReducer,
 	);
 
+	const validatedReducer: Reducer<
+		(RootState & PersistPartial) | undefined
+	> = (state, action) => {
+		if (action.type === 'persist/REHYDRATE') {
+			const decoded = persistedStateCodecNew.decode(action.payload);
```
grzpab, line 24: doesn't it return a validations structure?

grzpab (approved):

PR #536, opened 2023-06-19
fix

DmytroHryshyn:


intuita-webview/src/main/App.tsx
```diff
@@ -46,7 +46,6 @@ const PANELS: {
 		id: 'communityView',
 		title: 'Community',
 		Component: CommunityView,
-		inlineStyle: { flex: 'auto' },
```
grzpab, line 49: this was added by @hbjORbj, can you check with him?

PR #540, opened 2023-06-19
Dmytro/int 1213 some codemod runs are not persisted

DmytroHryshyn:


src/data/index.ts
```diff
@@ -12,7 +12,7 @@ const buildStore = (workspaceState: Memento) => {
 		{
 			key: 'compressedRoot',
 			storage: new MementoStorage(workspaceState),
-			throttle: 1000,
+			// throttle: 1000,
```
grzpab, line 15: @DmytroHryshyn as agreed, you'll fix that in the next PR, right?

grzpab (approved):

PR #541, opened 2023-06-19
INT-1206 implement the selectExplorerTree selector

grzpab:


DmytroHryshyn (approved):

src/selectors/selectExplorerTree.ts
```diff
@@ -0,0 +1,223 @@
...
+	rootNodeHashDigest: ExplorerNodeHashDigest;
+	nodes: Record<ExplorerNodeHashDigest, ExplorerNode>;
+	children: Record<
+		ExplorerNodeHashDigest,
+		ReadonlyArray<ExplorerNodeHashDigest>
+	>;
+	parents: Record<ExplorerNodeHashDigest, ExplorerNodeHashDigest>;
+	selectedNodeHashDigest: ExplorerNodeHashDigest | null;
+	expandedNodeHashDigests: ReadonlyArray<ExplorerNodeHashDigest>;
+	appliedJobHashes: ReadonlyArray<JobHash>;
```
DmytroHryshyn, line 98: do we want to keep domain specific field `appliedJobHashes` in tree structure or move it to separate structure?

DmytroHryshyn: maybe lets rename it to something like `selectedTreeNodes` (common field for all trees that have checkboxes and can be selected) or move it to other structure

src/selectors/selectExplorerTree.ts
```diff
@@ -0,0 +1,223 @@
...
+			hashDigest: ExplorerNodeHashDigest;
+			kind: 'DIRECTORY';
+			label: string;
+	  }>
+	| Readonly<{
+			hashDigest: ExplorerNodeHashDigest;
+			kind: 'FILE';
+			label: string;
+			jobHash: string;
+			fileAdded: boolean;
```
DmytroHryshyn, line 43: do we need `fileAdded` and `jobHash` fields directly on the Node object? can we somehow make  ExplorerNode  objects to have the same shape? otherwise it can impact performance.

src/selectors/selectExplorerTree.ts
grzpab: selectedTreeNodes are already a used name (selectedNodeHashDigest), the concept of selection (focus?) is the concept for the tree

src/selectors/selectExplorerTree.ts
grzpab: the webview needs to know when to render a file added icon or a normal icon based on the job kind (which we calculate as `fileAdded`)

src/selectors/selectExplorerTree.ts
DmytroHryshyn: maybe then lets rename `selectedNodeHashDigest ` to `focusedNodeHashDigest` and use name `selectedNodeHashDigests` to store nodes that are selected using checkboxes?

PR #545, opened 2023-06-20
Dynamic UI for Tree Item in Discovery panel

hbjORbj:
### Changes

- on narrow screen, let's still show dry-run button and edit-path button on hover
- on narrow screen, let's still show label
- on narrow screen, let's use only icons for dry-run button and edit-path button

### DIFF
--- a/intuita-webview/src/codemodList/TreeView/index.tsx
+++ b/intuita-webview/src/codemodList/TreeView/index.tsx
@@ -253,7 +253,7 @@ const TreeView = ({
 	}, [state]);
 
 	useEffect(() => {
-		if (typeof ResizeObserver === 'undefined') {
+		if (ResizeObserver === undefined) {
 			return undefined;
 		}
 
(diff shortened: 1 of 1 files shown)


intuita-webview/src/codemodList/TreeView/index.tsx
```diff
@@ -249,6 +252,40 @@ const TreeView = ({
 		});
 	}, [state]);
 
+	useEffect(() => {
+		if (typeof ResizeObserver === 'undefined') {
```
grzpab, line 256: why not a direct check against undefined?

intuita-webview/src/codemodList/TreeView/index.tsx
```diff
@@ -249,6 +252,40 @@ const TreeView = ({
...
+			} = container;
+
+			setTreeContainerWidth(width);
+		});
+
+		resizeObserver.observe(discoveryViewContainer);
+
+		return () => {
+			if (resizeObserver) {
+				resizeObserver.disconnect();
```
grzpab, line 284: `resizeObserver?.disconnect();` and `if` will no longer be necessary

grzpab (approved):

PR #549, opened 2023-06-20
INT-1196: Merge Diff Panel and Documentation View into one web view

hbjORbj:
#### Linear: https://linear.app/intuita/issue/INT-1196/merge-diff-panel-and-documentation-view-into-one-web-view

### DIFF
--- a/src/components/webview/IntuitaPanelProvider.ts
+++ b/src/components/webview/IntuitaPanelProvider.ts
@@ -47,7 +47,7 @@ const selectPanelViewProps = (
 	}
 
 	const validation1 = codemodMetadataCodec.decode(mainPanel);
-	console.log(validation1, 'VALIDATION 1');
+
 	if (validation1._tag === 'Right') {
 		const { title, doc } = validation1.right;
 		return {
@@ -58,7 +58,7 @@ const selectPanelViewProps = (
 	}
 
 	const validation2 = jobViewCodec.decode(mainPanel);
-	console.log(validation2, 'VALIDATION 2');
+
 	if (validation2._tag === 'Right') {
 		const { selectedCaseHash, focusedJobHash } = validation2.right;
 


DmytroHryshyn (approved):

intuita-webview/src/campaignManager/App.tsx
```diff
@@ -22,7 +22,7 @@ const handleItemClick = (node: CaseTreeNode) => {
 	});
 };
 
-type Props = { screenWidth: number | null };
```
grzpab, line 25: why this was changed to undefined? null is way better for checks

intuita-webview/src/jobDiffView/App.tsx
```diff
@@ -97,6 +105,19 @@ function App() {
 		return null;
 	}
 
+	if (view.metadataMarkdown !== null) {
```
grzpab, line 108: I think that you should pass `kind` within the `viewProps` to signify the kind of the view

intuita-webview/src/jobDiffView/App.tsx
```diff
@@ -28,8 +29,12 @@ export const App = () => {
 	}, []);
 
 	if (viewProps.kind === 'CODEMOD') {
-		// a placeholder for the codemod description (Markdown)
-		return null;
+		return viewProps.docs !== null ? (
```
grzpab, line 32: One idea about nullability that I have coined but probably haven't shared (:sweat_smile:) is to lock out of nullability as soon as possible.
In this case, `viewProps.docs` should not be null, if there's nothing to render, do not render anything (or display an info/error, if applicable).
The fewer nullables we have to handle, the better.

src/persistedState/codecs.ts
```diff
@@ -107,6 +107,18 @@ export const persistedStateCodecNew = buildTypeCodec({
 			selectedCaseHash: null,
 		},
 	),
+	panelView: withFallback(
+		buildTypeCodec({
```
grzpab, line 111: this should be a union codec representing the following states:
* a meaningful codemod description to render
* a meaningful job to render
* nothing to render (meaning the panel will never be displayed); default value for `panel`

with such a union our need to deal with nullability is drastically reduced

src/persistedState/codecs.ts
```diff
@@ -107,6 +107,18 @@ export const persistedStateCodecNew = buildTypeCodec({
 			selectedCaseHash: null,
 		},
 	),
+	panelView: withFallback(
```
grzpab, line 110: could be just `panel`

src/extension.ts
```diff
@@ -124,13 +124,20 @@ export async function activate(context: vscode.ExtensionContext) {
...
 					}
 
-					vscode.commands.executeCommand('markdown.showPreview', uri);
+					const metadataContent =
+						textContentProvider.provideTextDocumentContent(uri);
+
+					store.dispatch(
+						actions.setPanelView({
+							kind: 'CODEMOD',
+							title: `${name}.md`,
```
grzpab, line 137: we can provide the codemod name as the title, we definitely don't want the `.md` to be visible (we needed it only for the native markdown panel to work)

src/persistedState/codecs.ts
```diff
@@ -107,6 +107,18 @@ export const persistedStateCodecNew = buildTypeCodec({
...
+		buildTypeCodec({
+			kind: t.union([t.literal('JOB'), t.literal('CODEMOD'), t.null]),
+			docs: t.union([t.string, t.null]),
+			title: t.string,
```
grzpab, line 114: actually, for the `JOB`, the title will be calculated from other state properties.

src/persistedState/codecs.ts
grzpab: you could probably store the `codemodHash` instead of `docs` to point to the codemod you wanna display. The selector should take care of extracting the information.

grzpab (changes requested):
This PR was superseded by another one here: https://github.com/intuita-inc/intuita-vscode-extension/pull/598

PR #558, opened 2023-06-21
INT-1224: Store/Use Panel Tab id properly

hbjORbj:
#### Linear: https://linear.app/intuita/issue/INT-1224/storeuse-panel-tab-id-properly

### DIFF
--- a/intuita-webview/src/main/App.tsx
+++ b/intuita-webview/src/main/App.tsx
@@ -68,7 +68,7 @@ function App() {
 
 	const handlePanelTabClick = (id: TabKind) => {
 		vscode.postMessage({
-			kind: 'webview.main.setState',
+			kind: 'webview.main.setActiveTabId',
 			activeTabId: id,
 		});
 	};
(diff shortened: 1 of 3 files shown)


src/components/webview/webviewEvents.ts
```diff
@@ -293,6 +293,10 @@ export type WebviewResponse =
 			openedIds: ReadonlyArray<TreeNodeId>;
 			focusedId: TreeNodeId | null;
 	  }>
+	| Readonly<{
+			kind: 'webview.main.setState';
```
grzpab, line 297: `'webview.main.setActiveTabId'`

intuita-webview/src/main/App.tsx
```diff
@@ -65,19 +66,44 @@ function App() {
...
+						handlePanelTabClick(TabKind.codemods);
+					}}
+				>
 					Codemods Discovery
 				</VSCodePanelTab>
-				<VSCodePanelTab className="vscode-tab" id={TabKind.codemodRuns}>
+				<VSCodePanelTab
+					className="vscode-tab"
+					id={TabKind.codemodRuns}
+					onClick={() => {
```
DmytroHryshyn, line 94: can you check if we  can use `activeidChanged` callback on `VSCodePanels` component instead of adding click handlers to each tab?

PR #563, opened 2023-06-21
add codemod node renderer

DmytroHryshyn:
-adds codemod node renderer

### DIFF
--- a/src/selectors/selectCodemodTree.ts
+++ b/src/selectors/selectCodemodTree.ts
@@ -120,11 +120,11 @@ export const selectCodemodTree = (state: RootState) => {
 		}
 
 		const expanded =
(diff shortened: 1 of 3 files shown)


src/selectors/selectCodemodTree.ts
```diff
@@ -98,12 +107,56 @@ export const selectCodemodTree = (state: RootState) => {
...
+		depth: number,
+	) => {
+		const node = nodes[hashDigest] ?? null;
+
+		if (node === null) {
+			return;
+		}
+
+		const expanded =
+			state.changeExplorerView.openedFileExplorerNodeIds.includes(
```
grzpab, line 123: this won't work, you need to pick the nodes from the codemod view

grzpab (approved):
I assume this will change anyway after you connect it to the `IntuitaTreeView`.

PR #573, opened 2023-06-22
INT-1229 make the community tab stateless

grzpab:
Changes:
* remove the `ExternalLink` type (not needed)
* move the external links directly to the `CommunityTab` component
* remove the `index.tsx` and `index.html` files from the previous `CommunityView` (doesn't exist)
* make the `Community` class stateless (soon we will remove it anyway)
* no events will be sent to the `CommunityTab` from the extension


src/components/webview/CommunityProvider.ts
```diff
@@ -1,65 +1,17 @@
...
-	}
-
-	private __onDidReceiveMessage = (message: WebviewResponse) => {
-		if (message.kind === 'webview.command') {
-			if (message.value.command === 'openLink') {
-				commands.executeCommand(
-					'vscode.open',
-					Uri.parse(message.value.arguments?.[0] ?? ''),
-				);
+		webviewView.webview.onDidReceiveMessage((message: WebviewResponse) => {
```
DmytroHryshyn, line 6: I think we dont need this  command anymore, because we now use `VSCodeLink `native components in webview. 
I think that it redirect user to the given url directly (But this needs to be checked).

grzpab: lemme check, if it works, it then saves us even more code lines.

DmytroHryshyn (approved):

src/components/webview/CommunityProvider.ts
grzpab: (it did work, removing more code)

PR #574, opened 2023-06-22
Dmytro/int 1223 migrate codemodtree to the new tree structure 4

DmytroHryshyn:


intuita-webview/src/codemodList/TreeView/index2.tsx
```diff
@@ -0,0 +1,235 @@
...
+				kind: 'webview.codemodList.dryRunCodemod' as const,
+				value: node.hashDigest as unknown as CodemodHash,
+			},
+		];
+
+		return actionButtons(node, actions, doesDisplayShortenedTitle);
+	};
+
+	return (
+		<div id="codemodDiscoveryView-treeContainer" tabIndex={0}>
```
grzpab, line 211: actually we don't need those `div` with `id` and `tabIndex`, the `IntuitaTreeView` handles that

intuita-webview/src/codemodList/TreeView/index2.tsx
```diff
@@ -0,0 +1,235 @@
...
+import { getCodemodNodeRenderer } from '../codemodNodeRenderer';
+import Popover from '../../shared/Popover';
+import { VSCodeButton } from '@vscode/webview-ui-toolkit/react';
+
+import s from './style.module.css';
+
+type Props = Readonly<{
+	tree: CodemodTree;
+	autocompleteItems: string[];
+	searchQuery: string;
```
grzpab, line 22: IMHO we should call it `searchPhrase` to have the same names for the same concept everywhere (and search query is mostly associated with URLs I think)

grzpab (approved):

PR #577, opened 2023-06-22
Dmytro/int 1223 migrate codemodtree to the new tree structure 5

DmytroHryshyn:


intuita-webview/src/codemodList/CodemodNodeRenderer/index.tsx
```diff
@@ -182,6 +168,42 @@ const getCodemodNodeRenderer =
 			setEditingPath(false);
 		}, []);
 
+		// @TODO check if can be implemented using css container queries
```
grzpab, line 171: that's a nice idea

grzpab (approved):

PR #581, opened 2023-06-22
INT-1206 remove the index.html and index.tsx files from unused view

grzpab:


intuita-webview/src/main/CodemodRuns.tsx
```diff
@@ -35,7 +35,7 @@ const RESIZABLE_PANELS: {
 				command: 'intuita.clearState',
 			},
 		],
-		Component: (props) => CampaignManager(props),
+		Component: () => CampaignManager(),
```
DmytroHryshyn, line 38: i think it can be just `Component: CampaignManager`

DmytroHryshyn (approved):

PR #588, opened 2023-06-23
fix initial state

DmytroHryshyn:


src/data/index.ts
```diff
@@ -7,11 +7,32 @@ import { Memento } from 'vscode';
...
+	const parsedState: Record<string, unknown> = {};
+
+	try {
+		const rawState = JSON.parse(serializedState);
+
+		Object.keys(rawState).forEach((key) => {
+			parsedState[key] = JSON.parse(rawState[key]);
+		});
+	} catch (e) {
+		console.log(e);
```
grzpab, line 23: `console.error`

src/data/index.ts
```diff
@@ -7,11 +7,32 @@ import { Memento } from 'vscode';
...
+const PERSISTANCE_KEY = 'compressedRoot';
+
+const deserializeState = (serializedState: string) => {
+	const parsedState: Record<string, unknown> = {};
+
+	try {
+		const rawState = JSON.parse(serializedState);
+
+		Object.keys(rawState).forEach((key) => {
+			parsedState[key] = JSON.parse(rawState[key]);
```
grzpab, line 20: hmm, can you explain why that's needed?

src/data/index.ts
```diff
@@ -35,12 +56,26 @@ const buildStore = (workspaceState: Memento) => {
...
+	const deserializedState = deserializeState(initialState);
+
+	const decodedState = persistedStateCodecNew.decode(deserializedState);
+
+	// should never happen because of codec fallback
+	if (decodedState._tag !== 'Right') {
+		throw new Error('Invalid state');
+	}
+
+	console.log(decodedState.right);
```
grzpab, line 71: can you remove this line?

src/data/index.ts
```diff
@@ -7,11 +7,32 @@ import { Memento } from 'vscode';
...
+		console.log(e);
+	}
+
+	return parsedState;
+};
+
+const buildStore = async (workspaceState: Memento) => {
+	const storage = new MementoStorage(workspaceState);
+
 	const persistedReducer = persistReducer(
```
grzpab, line 32: do we need the `persistReducer` anymore? only for writes, am I right?

src/data/index.ts
DmytroHryshyn: this is needed because `redux-persist` stores each property as a string, so you need to parse the state object, and then parse each field

DmytroHryshyn: link: https://github.com/rt2zz/redux-persist/blob/d8b01a085e3679db43503a3858e8d4759d6f22fa/src/getStoredState.ts#L28

src/data/index.ts
DmytroHryshyn: we still need it, and we still need hydration action to be fired, otherwise persisting does not work. i think its because it relies on `persisted` flag, but i need to check it. in future we can think about removing this library at all and just subscribe to store and write directly to memento when soemthing chnages.

src/data/index.ts
grzpab: @DmytroHryshyn does it mean we need to do this recursively, or just one level?

DmytroHryshyn: only one level

grzpab (approved):

PR #593, opened 2023-06-23
split renderer

DmytroHryshyn:
-separate nodeRenderer component to 2 components: Codemod and Directory 
-removes a lot of conditions, improves readability, increase performance because resizing hook and logic is now only within Codemod component, and Directory is small and memoized


intuita-webview/src/codemodList/CodemodNodeRenderer/Codemod.tsx
```diff
@@ -0,0 +1,214 @@
...
+	});
+}, 50);
+
+type Props = Readonly<{
+	nodeDatum: NodeDatum<
+		CodemodNodeHashDigest,
+		CodemodNode & { kind: 'CODEMOD' }
+	>;
+	rootPath: string;
+	autocompleteItems: string[];
```
grzpab, line 41: could be `ReadonlyArray<string>`

PR #600, opened 2023-06-26
INT-1241 tree-item focus change

grzpab:
Changes:
* do not use focus within tree-item but use scrollIntoView (just like in the codemod tree item)
* remove non-existing buttons (essentially  configuration) from package.json

### DIFF
--- a/intuita-webview/src/shared/TreeItem/index.tsx
+++ b/intuita-webview/src/shared/TreeItem/index.tsx
@@ -41,8 +41,20 @@ const TreeItem = ({
 
 	useLayoutEffect(() => {
 		if (focused) {
-			ref.current?.focus();
+			const timeout = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


intuita-webview/src/shared/TreeItem/index.tsx
```diff
@@ -41,8 +41,20 @@ const TreeItem = ({
...
 	useLayoutEffect(() => {
 		if (focused) {
-			ref.current?.focus();
+			const timeout = setTimeout(() => {
+				ref.current?.scrollIntoView({
+					behavior: 'smooth',
+					block: 'center',
+					inline: 'center',
+				});
+			}, 0);
```
DmytroHryshyn, line 50: we dont need delay anymore, right?

DmytroHryshyn (approved):

PR #628, opened 2023-06-27
INT-1246 ensure the main provider has only one props builder

grzpab:


intuita-webview/src/codemodList/App.tsx
```diff
@@ -13,55 +12,19 @@ const setSearchPhrase = (searchPhrase: string) => {
...
-			</main>
-		);
-	}
-
-	const { codemodTree, autocompleteItems, searchPhrase, rootPath } =
-		view.viewProps;
-
-	return (
+export const App = memo(
+	(props: MainWebviewViewProps & { activeTabId: 'codemods' }) => (
```
DmytroHryshyn, line 16: if we pass props to the App component `memo` will not work without comparator, because props are nested object

DmytroHryshyn (approved):

PR #649, opened 2023-06-29
Feature/open panel only when clicked

DmytroHryshyn:


DmytroHryshyn:
> can you explain what this is about?

currently we have a bug: job diff panel is reopened each time user update extension state. we decided that job diff panel should reopen only if user clicks on file/case or clicks on codemod.

PR #661, opened 2023-06-30
INT-1280 Refactor Intuita Popper

hbjORbj:


intuita-webview/src/jobDiffView/DiffViewer/Header/index.tsx
```diff
@@ -33,10 +33,10 @@ export const Header = (props: Props) => {
 							/>
 						</VSCodeButton>
 					}
-					popoverText="Move to the previous file"
+					content="Move to the previous file"
 				/>
-				<Popover
-					trigger={
+				<IntuitaPopover
```
DmytroHryshyn, line 38: I think it can be more readable if we wrap children with IntuitaPopover like this: 
```
<IntuitaPopover> 
  <VsCodeButton />
  </IntuitaPopover>
  ```
  instead of passing `children` prop. 1-st way feels more natural.

DmytroHryshyn (approved):

PR #668, opened 2023-07-03
Dmytro/int 1296 fileexplorer search fix and make similar to codesee

DmytroHryshyn:
https://linear.app/intuita/issue/INT-1296/fileexplorer-search-fix-and-make-similar-to-codesee

### DIFF
--- a/src/components/webview/MainProvider.ts
+++ b/src/components/webview/MainProvider.ts
@@ -241,6 +241,13 @@ export class MainViewProvider implements WebviewViewProvider {
 					message.searchPhrase,
 				]),
 			);
+
+			this.__store.dispatch(
+				actions.setExplorerNodes([
+					message.caseHashDigest,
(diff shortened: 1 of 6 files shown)


src/components/webview/MainProvider.ts
```diff
@@ -241,6 +241,13 @@ export class MainViewProvider implements WebviewViewProvider {
 					message.searchPhrase,
 				]),
 			);
+
+			this.__store.dispatch(
```
grzpab, line 245: why is this action needed?

DmytroHryshyn: to recalculate explorerNodes. i've tried to put search logic to the selector where it should be, but had a problem that i cannot keep correct tree structure this way. so i had to put this logic to reducer.

grzpab (approved):

PR #672, opened 2023-07-03
remove unused job fields

DmytroHryshyn:
- removes, oldDataPath, oldContentUri

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -564,10 +564,6 @@ export async function activate(context: vscode.ExtensionContext) {
 				if (job.newContentUri !== null) {
 					uris.push(vscode.Uri.parse(job.newContentUri));
 				}
-
-				if (job.oldContentUri !== null) {
-					uris.push(vscode.Uri.parse(job.oldContentUri));
-				}
 			}
 
 			store.dispatch(actions.clearState());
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -547,15 +546,17 @@ export async function activate(context: vscode.ExtensionContext) {
...
+				if (!job) {
 					continue;
 				}
 
-				uris.push(vscode.Uri.parse(job.newContentUri));
+				if (job.newContentUri !== null) {
+					uris.push(vscode.Uri.parse(job.newContentUri));
+				}
+
+				if (job.oldContentUri !== null) {
```
grzpab, line 557: I think this `oldContentUri` is the same as `oldUri` right now, because we don't preformat files right now.

src/extension.ts
```diff
@@ -546,15 +545,13 @@ export async function activate(context: vscode.ExtensionContext) {
 			const uris: vscode.Uri[] = [];
 
 			for (const job of Object.values(state.job.entities)) {
-				if (
-					!job ||
-					!doesJobAddNewFile(job.kind) ||
```
grzpab, line 551: I think this part should be restored :)

grzpab (approved):

PR #677, opened 2023-07-04
INT-1255 Revamp Checkbox state

hbjORbj:
Changes:
- If some children of a folder are deselected, use Indeterminate icon.
- If all children of a folder are deselected, the folder must have blank checkbox
- If all children of a folder are selected, the folder must have checked checkbox

https://github.com/intuita-inc/intuita-vscode-extension/assets/32841130/02754720-9c4b-4340-a9ac-bdbfd54c02f1

### DIFF
--- a/src/data/slice.ts
+++ b/src/data/slice.ts
@@ -81,7 +81,9 @@ export const getInitialState = (): RootState => {
 		explorerSearchPhrases: {},
 		explorerNodes: {},
 		selectedExplorerNodes: {},
-		explorerNodesWithDeselectedChildNodes: {},
+		// indeterminate explorer node is a node some (but not all) of whose children are deselected.
+		// For such node, we will use indeterminate checkbox icon.
+		indeterminateExplorerNodes: {},
 		collapsedExplorerNodes: {},
 		focusedExplorerNodes: {},
 	};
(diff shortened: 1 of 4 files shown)


src/data/slice.ts
```diff
@@ -492,45 +502,142 @@ const rootSlice = createSlice({
...
+					}
+				}
+
+				const allSelected =
+					childNodeHashDigests.length > 0 &&
+					childNodeHashDigests.every((hashDigest) =>
+						updatedSelectedHashDigests.includes(hashDigest),
+					);
+				const allDeselected =
+					childNodeHashDigests.length > 0 &&
```
grzpab, line 595: the length check is not needed (every for an empty array is true)

src/data/slice.ts
```diff
@@ -492,45 +502,142 @@ const rootSlice = createSlice({
...
+						break;
+					}
+
+					if (node.depth === explorerNode.depth) {
+						childNodeHashDigests.push(node.hashDigest);
+					}
+				}
+
+				const allSelected =
+					childNodeHashDigests.length > 0 &&
```
grzpab, line 590: the length check is not needed (every for an empty array is true)

grzpab (approved):

PR #682, opened 2023-07-04
Diff view editable

DmytroHryshyn:


intuita-webview/src/jobDiffView/DiffViewer/Diff.tsx
```diff
@@ -2,62 +2,96 @@ import { memo, useEffect, useRef } from 'react';
 import MonacoDiffEditor from '../../shared/Snippet/DiffEditor';
 import { getDiff, Diff } from '../../shared/Snippet/calculateDiff';
 import type { editor } from 'monaco-editor';
+import { Disposable } from 'vscode';
 
 export type { Diff };
+type Props = {
```
grzpab, line 8: can you make the Props readonly?

src/components/webview/webviewEvents.ts
```diff
@@ -162,4 +162,9 @@ export type WebviewResponse =
 	| Readonly<{
 			kind: 'webview.global.flipCodemodHashDigest';
 			codemodNodeHashDigest: CodemodNodeHashDigest;
+	  }>
+	| Readonly<{
+			kind: 'webview.jobDiffView.contentModified';
```
grzpab, line 167: I think we should call `jobDiffView` just `panel` because it shares job diffing and codemod description.

grzpab (approved):

PR #691, opened 2023-07-05
Move explorer nodes to selector

DmytroHryshyn:


grzpab (approved):
@DmytroHryshyn @hbjORbj please do take into account that this PR will make the state calculations slightly slower.

PR #693, opened 2023-07-05
Dmytro/int 1279 add ability to apply case partially

DmytroHryshyn:


intuita-webview/src/fileExplorer/App.tsx
```diff
@@ -121,7 +121,7 @@ export const App = (
 				<ActionsFooter
 					caseHash={changeExplorerTree.caseHash}
 					screenWidth={props.screenWidth}
-					selectedJobCount={changeExplorerTree.selectedJobCount}
```
DmytroHryshyn, line 124: Can we keep `{changeExplorerTree.selectedJobCount}` here?

grzpab (approved):

PR #699, opened 2023-07-06
INT-1313 add ability to run Piranha rules

grzpab:
Changes:
* added the `intuita.executeAsPiranhaRule` command,
* added the `codemodEngineNodeExecutableUri` (now hardcoded) URI,
* removed unused and doubled property from the `Execution` type,
* take the `codemodName` from the state or from the command

### DIFF
--- a/src/components/buildArguments.ts
+++ b/src/components/buildArguments.ts
@@ -1,0 +1,100 @@
+import { Uri } from 'vscode';
+import type { Configuration } from '../configuration';
+import type { Message, MessageKind } from './messageBus';
+import { doubleQuotify, singleQuotify } from '../utilities';
+
+export const buildArguments = (
+	configuration: Configuration,
+	message: Omit<
+		Message & { kind: MessageKind.executeCodemodSet },
+		'storageUri'
+	>,
+	storageUri: Uri,
+) => {
+	const args: string[] = [];
+
+	const { command } = message;
+
+	if (command.kind === 'executePiranhaRule') {
+		args.push('-i', singleQuotify(message.targetUri.fsPath));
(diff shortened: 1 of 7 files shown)


src/components/buildArguments.ts
```diff
@@ -0,0 +1,100 @@
...
+	const args: string[] = [];
+
+	const { command } = message;
+
+	if (command.kind === 'executePiranhaRule') {
+		args.push('-i', singleQuotify(message.targetUri.fsPath));
+		args.push('-c', singleQuotify(command.configurationUri.fsPath));
+		args.push('-o', singleQuotify(storageUri.fsPath));
+
+		// configuration.includePatterns.forEach((includePattern) => {
```
grzpab, line 23: I need to figure out why this doesn't work, but this is out of the scope of this PR.

src/extension.ts
```diff
@@ -286,6 +286,52 @@ export async function activate(context: vscode.ExtensionContext) {
...
+			'intuita.executeAsPiranhaRule',
+			async (configurationUri: vscode.Uri) => {
+				const fileStat = await vscode.workspace.fs.stat(
+					configurationUri,
+				);
+				const configurationUriIsDirectory = Boolean(
+					fileStat.type & vscode.FileType.Directory,
+				);
+
+				if (!configurationUriIsDirectory) {
```
DmytroHryshyn, line 300: do we need this check? this command is only available in right-click menu for directories, right?

grzpab: I would say better safe than sorry :) We can accidentally remove the restriction from `package.json` settings.

DmytroHryshyn (approved):

PR #723, opened 2023-07-11
No need to use file path for directory node

hbjORbj:


src/selectors/selectExplorerTree.ts
```diff
@@ -115,7 +115,7 @@ export const selectExplorerNodes = (
 
 				return {
 					kind: 'DIRECTORY' as const,
-					path,
+					path: directoryPath,
```
grzpab, line 118: can you explain this change? what was the issue?

grzpab (approved):

PR #724, opened 2023-07-11
INT-1301 all folders must come before all files given that they are at the same depth

hbjORbj:
### Results

Screenshot 2023-07-11 at 11 00 25
Screenshot 2023-07-11 at 11 00 30
Screenshot 2023-07-11 at 11 00 40

### DIFF
--- a/src/selectors/selectExplorerTree.ts
+++ b/src/selectors/selectExplorerTree.ts
@@ -191,10 +191,13 @@ export const selectExplorerNodes = (
 			(hashA, hashB) => {
 				const childA = nodes[hashA];
 				const childB = nodes[hashB];
-				if (childA?.kind === 'DIRECTORY' && childB?.kind === 'FILE') {
+				if (!childA || !childB) {
+					return 0;
+				}
+				if (childA.kind === 'DIRECTORY' && childB.kind === 'FILE') {
 					return -1;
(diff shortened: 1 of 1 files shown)


src/selectors/selectExplorerTree.ts
```diff
@@ -186,7 +186,22 @@ export const selectExplorerNodes = (
 
 		node = { ...node, depth };
 
-		selectedFileCount = Array.from(children[node.hashDigest] ?? [])
+		// all folders must come before all files given that they are at the same depth
+		const sortedChildren = Array.from(children[node.hashDigest] ?? []).sort(
+			(hashA, hashB) => {
+				const childA = nodes[hashA];
+				const childB = nodes[hashB];
```
grzpab, line 193: return 0 if either child does not exist

intuita-webview/src/codemodList/CodemodNodeRenderer/index.tsx
```diff
@@ -53,6 +53,7 @@ const getCodemodNodeRenderer =
 				block: 'center',
 				inline: 'center',
 			});
+			ref.current?.focus();
```
grzpab, line 56: are we sure we do want to focus in here?

PR #745, opened 2023-07-18
INT-1389 Users should be able to browse Intuita extension (with codemod action items disabled) even if no workspace is chosen

hbjORbj:
https://app.claap.io/intuita/int-1389-users-should-be-able-to-browse-intuita-extension-with-codemod-action-items-disabled-even-if-no-workspace-is-chosen-c-BT2DRbCjzO-Dbxy2S3zIrq-

### DIFF
--- a/src/components/webview/IntuitaPanelProvider.ts
+++ b/src/components/webview/IntuitaPanelProvider.ts
@@ -76,11 +76,7 @@ const selectPanelViewProps = (
 		};
 	}
 
-	if (rootPath === null) {
-		return null;
-	}
-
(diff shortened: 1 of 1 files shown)


src/components/webview/IntuitaPanelProvider.ts
```diff
@@ -80,6 +76,14 @@ const selectPanelViewProps = (
 		};
 	}
 
+	if (rootPath === null) {
+		return null;
+	}
+
+	if (activeTabId === 'community') {
```
grzpab, line 83: this could be joined with the statement on line 79

PR #752, opened 2023-07-21
INT-1408 [UX] Initial state of codemod directories should be collapsed

hbjORbj:


grzpab (approved):
@hbjORbj please test it thoroughly

PR #756, opened 2023-07-24
INT-1433 Allow private codemod sharing via permalink in studio

hbjORbj:
relevant PR: https://github.com/intuita-inc/studio/pull/232

### DIFF
--- a/src/components/engineService.ts
+++ b/src/components/engineService.ts
@@ -20,8 +20,9 @@ import { CodemodHash } from '../packageJsonAnalyzer/types';
 import { ExecutionError, executionErrorCodec } from '../errors/types';
 import {
 	CodemodEntry,
-	codemodEntryCodec,
 	codemodNamesCodec,
+	PrivateCodemodEntry,
+	privateCodemodEntryCodec,
 } from '../codemods/types';
 import { actions } from '../data/slice';
 import { Store } from '../data';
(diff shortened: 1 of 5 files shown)


src/components/engineService.ts
```diff
@@ -200,7 +200,9 @@ export class EngineService {
 
 	public async fetchPrivateCodemods(): Promise<void> {
 		try {
-			const privateCodemods: CodemodEntry[] = [];
+			const privateCodemods: (CodemodEntry & {
+				permalink: string | null;
```
grzpab, line 204: can't we calculate the permalink on demand? IMO this shouldn't be part of the entry

src/extension.ts
```diff
@@ -874,12 +881,20 @@ export async function activate(context: vscode.ExtensionContext) {
 						);
 					}
 					newPrivateCodemodNames.push(codemodHash);
-					await writeFile(
-						privateCodemodNamesPath,
-						JSON.stringify({
-							names: newPrivateCodemodNames,
-						}),
-					);
+					await Promise.all([
```
grzpab, line 884: why do we need Promise all here?

grzpab: I think it's negligent with two promises, it's easier to read it when it's one by one (but it's ok to leave it as it is)

PR #785, opened 2023-09-01
INT-1752 Catch relevant top-level nodes for before/after snippets for reporting issue

hbjORbj:


package.json
```diff
@@ -247,9 +248,11 @@
 		"webpack-cli": "^5.0.1"
 	},
 	"dependencies": {
+		"@effect/schema": "0.30.3",
 		"@reduxjs/toolkit": "^1.9.5",
 		"@vscode/extension-telemetry": "^0.7.7",
 		"axios": "^1.2.2",
+		"diff": "^5.1.0",
```
grzpab, line 255: https://opensource.org/license/bsd-3-clause/

PR #796, opened 2023-09-07
INT-1752 Integration 3

hbjORbj:
- call `https://telemetry.intuita.io/sourceControl/github/issues` API request with title, body and repo
- after `html_url` is obtained, redirect user to the GH issue url
- type checking for API response

### DIFF
--- a/src/components/webview/SourceControlPanelProvider.ts
+++ b/src/components/webview/SourceControlPanelProvider.ts
@@ -220,7 +220,7 @@ export class SourceControlPanelProvider {
 							{
 								title,
 								body,
-								token: storedClerkToken,
+								accessToken: storedClerkToken,
 								repo: codemodRegistryRepo,
 							},
 						);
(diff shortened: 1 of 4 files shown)


src/components/webview/SourceControlPanelProvider.ts
```diff
@@ -209,8 +211,33 @@ export class SourceControlPanelProvider {
...
+						const { title, body } = message.data;
 						// call API to create Github Issue
+						const codemodRegistryRepo =
+							'https://github.com/intuita-inc/codemod-registry';
+						const { data } = await axios.post(
+							'https://telemetry.intuita.io/sourceControl/github/issues',
+							{
+								title,
+								body,
+								userId: storedClerkToken,
```
grzpab, line 223: hmm, is it a token, or an user ID? this shouldn't be confused.

hbjORbj: token. will change in studio backend

hbjORbj: https://github.com/intuita-inc/studio-backend/pull/46

hbjORbj: renamed it to `accessToken`. Ref: https://github.com/intuita-inc/studio-backend/pull/47

PR #798, opened 2023-09-08
Dmytro/int 1822 ive allow the user to pass arguments to any codemod in the

DmytroHryshyn:


src/components/buildArguments.ts
```diff
@@ -20,17 +20,10 @@ export const buildArguments = (
...
-
-		// 	args.push('-p', singleQuotify(fsPath));
-		// });
-
-		// configuration.excludePatterns.forEach((excludePattern) => {
-		// 	const { fsPath } = Uri.joinPath(message.targetUri, excludePattern);
-
-		// 	args.push('-a', singleQuotify(fsPath));
-		// });
+		const codemodArguments = (command.arguments ?? []).flatMap(
```
grzpab, line 23: if it's used in two places, I would calculate it before the split by `command.kind`

src/data/codemodConfigSchema.ts
```diff
@@ -1,27 +1,59 @@
...
+			name: S.string,
+			kind: S.literal('number'),
+			default: S.union(S.number, S.undefined),
+		}),
+		S.struct({
+			name: S.string,
+			kind: S.literal('boolean'),
+			default: S.union(S.boolean, S.undefined),
+		}),
+		// S.struct({
```
grzpab, line 20: should we keep it?

src/data/slice.ts
```diff
@@ -70,6 +70,8 @@ export const getInitialState = (): RootState => {
 			focusedCodemodHashDigest: null,
 			expandedNodeHashDigests: [],
 			searchPhrase: '',
+			codemodArgumentsPopupHashDigest: null,
```
grzpab, line 73: I think `codemodArguments` could be one structure, what do you think?

src/data/codemodConfigSchema.ts
DmytroHryshyn: i will uncomment it in  one of next pr when i add support of dropdowns

src/components/buildArguments.ts
```diff
@@ -12,31 +12,27 @@ export const buildArguments = (
...
+
+	const codemodArguments =
+		command.kind !== 'executeLocalCodemod'
+			? (command.arguments ?? []).flatMap(({ name, value }) => [
+					`--arg:${name}`,
+					String(value),
+			  ])
+			: [];
+
+	args.push(...codemodArguments);
```
grzpab, line 25: I think that won't work, these arguments must be passed as the last ones because the first argument must be the codemod name (the positional argument)

grzpab (approved):

PR #799, opened 2023-09-08
INT-1825 move create issue panel to the left as a tab + flow improvements

hbjORbj:
- [x] flow changes: after user comes back to VSCode from CS, automatically finish github creation (this means title and body must be stored in Redux) and show a toaster saying “Github issue created”
- [x] moving create issue panel to the left panel as a transient tab

Screenshot 2023-09-08 at 15 04 45


src/selectors/selectSourceControlTabProps.ts
```diff
@@ -0,0 +1,73 @@
...
+		<pre><code>${before ?? '// paste code here'}</code></pre>
+		<p><strong>2. Expected code after transformation (Desired output of codemod)</strong></p>
+		<pre><code>${expected ?? '// paste code here'}</code></pre>
+		<p><strong>3. Faulty code obtained after running the current version of the codemod (Actual output of codemod)</strong></p>
+		<pre><code>${after ?? '// paste code here'}</code></pre>
+		<h3>Additional context</h3>
+	`;
+};
+
+type SourceControlTabProps = {
```
grzpab, line 27: can you make it readonly?

src/selectors/selectMainWebviewViewProps.ts
```diff
@@ -55,6 +56,16 @@ export const selectMainWebviewViewProps = (
 		};
 	}
 
+	if (state.activeTabId === 'sourceControl') {
+		const sourceControlTabProps = selectSourceControlTabProps(state);
+
+		return {
+			activeTabId: state.activeTabId,
+			title: sourceControlTabProps?.title ?? null,
```
grzpab, line 64: I'd make it an empty string, or return `selectSourceControlTabProps` as the view props so we show what has happened in the app

grzpab (approved):
Approved, the minor issues can be fixed later.

src/selectors/selectMainWebviewViewProps.ts
hbjORbj: addressed

src/selectors/selectSourceControlTabProps.ts
hbjORbj: addressed

PR #801, opened 2023-09-11
INT-1835 pass any major language supported by Piranha

grzpab:


src/components/engineService.ts
```diff
@@ -343,7 +343,7 @@ export class EngineService {
 						kind: 'piranhaRule',
 						hashDigest,
 						name,
-						language: config.engine,
+						language: config.language,
```
grzpab, line 346: this was a bug

DmytroHryshyn (approved):

PR #809, opened 2023-09-13
INT-1853 Ensure that the syntax highlight of languages be rendered in GH

hbjORbj:
Screenshot 2023-09-13 at 07 58 04

claap: https://app.claap.io/intuita/ensure-that-the-syntax-highlight-of-languages-be-rendered-in-gh-c-BT2DRbCjzO-fKnJNUOwhgtp

### DIFF
--- a/intuita-webview/src/CreateIssue/index.tsx
+++ b/intuita-webview/src/CreateIssue/index.tsx
@@ -14,16 +14,12 @@ import './tiptap.css';
 
 const lowlight = createLowlight(common);
 
-const convertHTMLCodeBlockToMarkdownString = (htmlSnippet: string) => {
-	return htmlSnippet.replace(
+const convertHTMLCodeBlockToMarkdownString = (htmlSnippet: string) =>
+	htmlSnippet.replace(
 		/<pre><code(?:\s+class="language-(.*?)")?>(.*?)<\/code><\/pre>/gs,
-		function (_match, language, content) {
-			return `\n\n\`\`\`${
-				language ?? 'typescript'
-			}\n${content}\n\`\`\`\n\n`;
-		},
+		(_match, language, content) =>
+			`\n\n\`\`\`${language ?? 'typescript'}\n${content}\n\`\`\`\n\n`,
 	);
-};
 
 type Props = Readonly<{
 	title: string;
--- a/intuita-webview/src/CreateIssue/index.tsx
+++ b/intuita-webview/src/CreateIssue/index.tsx
@@ -14,6 +14,17 @@ import './tiptap.css';
 
 const lowlight = createLowlight(common);
 
+const convertHTMLCodeBlockToMarkdownString = (htmlSnippet: string) => {
+	return htmlSnippet.replace(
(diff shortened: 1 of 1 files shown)


intuita-webview/src/CreateIssue/index.tsx
```diff
@@ -14,6 +14,17 @@ import './tiptap.css';
 
 const lowlight = createLowlight(common);
 
+const convertHTMLCodeBlockToMarkdownString = (htmlSnippet: string) => {
+	return htmlSnippet.replace(
```
grzpab, line 18: you can remove the block and the return statement

intuita-webview/src/CreateIssue/index.tsx
```diff
@@ -14,6 +14,17 @@ import './tiptap.css';
...
+		/<pre><code(?:\s+class="language-(.*?)")?>(.*?)<\/code><\/pre>/gs,
+		function (_match, language, content) {
```
grzpab, line 20: this could be an arrow function I think

intuita-webview/src/CreateIssue/index.tsx
```diff
@@ -14,6 +14,17 @@ import './tiptap.css';
 
 const lowlight = createLowlight(common);
 
+const convertHTMLCodeBlockToMarkdownString = (htmlSnippet: string) => {
```
grzpab, line 17: why was that needed? aren't we controlling what HTML gets created?

intuita-webview/src/CreateIssue/index.tsx
```diff
@@ -78,8 +88,11 @@ const CreateIssue = (props: Props) => {
 	}, [props.title]);
 
 	useEffect(() => {
+		if (props.loading) {
+			return;
+		}
 		editor?.commands.setContent(props.body, false);
-	}, [editor, props.body]);
+	}, [editor, props.body, props.loading]);
```
grzpab, line 95: is this a bug fix?

intuita-webview/src/CreateIssue/index.tsx
hbjORbj: addressed

intuita-webview/src/CreateIssue/index.tsx
hbjORbj: addressed

intuita-webview/src/CreateIssue/index.tsx
hbjORbj: yes. this is needed to do the following check
```
if (props.loading) { return; }
```

which is needed to prevent the content changing when user presses "create issue" button. the content will change if `convertHTMLCodeBlockToMarkdownString` results in any change.

intuita-webview/src/CreateIssue/index.tsx
hbjORbj: The initial value we pass to tiptap editor is a HTML string; otherwise, we can't apply CSS styling.
GH doesn't handle `<pre><code class='language-typescript'>...</code></pre>`. Hence, we run this function before passing it to GH.

PR #815, opened 2023-09-18
INT-1845 Complete Auth UX with UI

hbjORbj:
claap: https://app.claap.io/intuita/int-1845-complete-auth-ux-with-ui-c-BT2DRbCjzO-DDxAN9mU0MtZ

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
 				"icon": "$(sign-in)"
 			},
 			{
-				"command": "intuita.signedIn",
+				"command": "intuita.handleSignedInUser",
 				"title": "Signed in",
 				"icon": "$(accounts-view-bar-icon)"
 			},
@@ -162,12 +162,12 @@
 			"view/title": [
 				{
 					"command": "intuita.signIn",
-					"when": "view == intuitaMainView && !intuita.signedIn",
+					"when": "view == intuitaMainView && !intuita.handleSignedInUser",
 					"group": "navigation"
 				},
 				{
-					"command": "intuita.signedIn",
-					"when": "view == intuitaMainView && intuita.signedIn",
+					"command": "intuita.handleSignedInUser",
+					"when": "view == intuitaMainView && intuita.handleSignedInUser",
 					"group": "navigation"
 				}
 			],
--- a/package.json
+++ b/package.json
@@ -39,6 +39,16 @@
 	"contributes": {
 		"commands": [
 			{
+				"command": "intuita.signIn",
+				"title": "Sign in",
+				"icon": "$(sign-in)"
+			},
+			{
+				"command": "intuita.signedIn",
+				"title": "Signed in",
+				"icon": "$(accounts-view-bar-icon)"
+			},
+			{
 				"command": "intuita.clearOutputFiles",
 				"title": "Intuita: clear output files"
 			},
@@ -149,6 +159,18 @@
 			]
 		},
 		"menus": {
+			"view/title": [
+				{
+					"command": "intuita.signIn",
+					"when": "view == intuitaMainView && !intuita.signedIn",
+					"group": "navigation"
+				},
+				{
+					"command": "intuita.signedIn",
+					"when": "view == intuitaMainView && intuita.signedIn",
+					"group": "navigation"
+				}
+			],
 			"commandPalette": [
 				{
 					"command": "intuita.clearState",
--- a/package.json
+++ b/package.json
@@ -162,12 +162,12 @@
 			"view/title": [
 				{
 					"command": "intuita.signIn",
-					"when": "view == intuitaMainView && !intuita.handleSignedInUser",
+					"when": "view == intuitaMainView && !intuita.signedIn",
 					"group": "navigation"
 				},
 				{
 					"command": "intuita.handleSignedInUser",
-					"when": "view == intuitaMainView && intuita.handleSignedInUser",
+					"when": "view == intuitaMainView && intuita.signedIn",
 					"group": "navigation"
 				}
 			],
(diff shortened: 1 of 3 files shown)


src/components/webview/MainProvider.ts
```diff
@@ -29,10 +30,36 @@ import {
...
 } from '../../selectors/selectCodemodTree';
-import { isNeitherNullNorUndefined } from '../../utilities';
+import { buildTypeCodec, isNeitherNullNorUndefined } from '../../utilities';
 
 const X_INTUITA_ACCESS_TOKEN = 'X-Intuita-Access-Token'.toLocaleLowerCase();
 
+export const validateAccessToken = async (
+	accessToken: string,
+): Promise<boolean> => {
+	const result = await axios.post(
```
DmytroHryshyn, line 40: can it throw an error? Do we need to wrap it in try/catch?

package.json
```diff
@@ -38,6 +38,16 @@
 	},
 	"contributes": {
 		"commands": [
+			{
+				"command": "intuita.signIn",
+				"title": "Sign in",
+				"icon": "$(sign-in)"
+			},
+			{
+				"command": "intuita.signedIn",
```
grzpab, line 47: can you explain the need for this command, from the name it sounds like a state fact

src/components/webview/MainProvider.ts
grzpab: yes, it could, but then it would anyway reject a promise

src/components/webview/MainProvider.ts
```diff
@@ -29,10 +30,36 @@ import {
...
+		'https://telemetry.intuita.io/validateAccessToken',
+		{},
+		{
+			headers: {
+				[X_INTUITA_ACCESS_TOKEN]: accessToken,
+			},
+		},
+	);
+	const { data } = result;
+	const validation = buildTypeCodec({
```
grzpab, line 50: like I mentioned in the backend part, we should use the http status code instead of validating the response body. In case of 4xx errors, Axios will reject the promise automatically.

src/extension.ts
```diff
@@ -71,6 +72,14 @@ export async function activate(context: vscode.ExtensionContext) {
 	);
 	const userService = new UserService(globalStateTokenStorage);
 
+	const accessToken = userService.getLinkedToken();
+
+	vscode.commands.executeCommand(
+		'setContext',
```
grzpab, line 78: can you explain this string literal here?

src/extension.ts
```diff
@@ -163,6 +172,41 @@ export async function activate(context: vscode.ExtensionContext) {
...
+
+			const url = new URL('https://codemod.studio');
+			url.search = searchParams.toString();
+
+			vscode.commands.executeCommand('intuita.redirect', url);
+		}),
+	);
+
+	context.subscriptions.push(
+		vscode.commands.registerCommand('intuita.signedIn', async () => {
```
grzpab, line 192: oh I see, I would call it with a verb, e.g. `intuita.handleSignedInUser` or something similar that conveys what kind of action we want to perform

package.json
hbjORbj: This is attached to a user icon that will appear on the right end of the screenshot attached only if user is already signed-in.

`command` is a required field. When pressing this, we will see a VSCode popup saying "You are signed in." with a button that says "Do you want to sign out?"

Screenshot 2023-09-18 at 10 26 50

src/extension.ts
hbjORbj: https://code.visualstudio.com/api/references/when-clause-contexts#add-a-custom-when-clause-context

src/extension.ts
hbjORbj: agree, addressed

src/components/webview/MainProvider.ts
hbjORbj: addressed

src/components/webview/MainProvider.ts
```diff
@@ -29,10 +30,40 @@ import {
...
+		'https://telemetry.intuita.io/validateAccessToken',
+		{},
+		{
+			headers: {
+				[X_INTUITA_ACCESS_TOKEN]: accessToken,
+			},
+		},
+	);
+	const { data, status } = result;
+	if (status !== 200) {
```
grzpab, line 50: I think now you don't need that, in general, if the promise rejects, you are not logged in (you may inspect the error code for 401 or other code to know what exactly the problem is).

src/components/webview/MainProvider.ts
```diff
@@ -29,10 +30,40 @@ import {
...
+export const validateAccessToken = async (
+	accessToken: string,
+): Promise<boolean> => {
+	const result = await axios.post(
+		'https://telemetry.intuita.io/validateAccessToken',
+		{},
+		{
+			headers: {
+				[X_INTUITA_ACCESS_TOKEN]: accessToken,
+			},
```
grzpab, line 46: please add a timeout to this request (I think 5s is the maximum time that makes sense here)>

src/extension.ts
grzpab: Nice, actually I wasn't aware of this feature.

src/components/webview/MainProvider.ts
hbjORbj: addressed

src/components/webview/MainProvider.ts
hbjORbj: addressed

src/components/webview/MainProvider.ts
```diff
@@ -29,10 +30,38 @@ import {
...
+			timeout: 5000,
+		},
+	);
+	const { data } = result;
+
+	const validation = buildTypeCodec({
```
grzpab, line 52: You don't need the body validation at all right now, right?

hbjORbj: addressed

PR #818, opened 2023-09-19
INT-1888 Show a settings icon to the left of the sign in/sign out icon

hbjORbj:
claap: https://app.claap.io/intuita/int-1888-show-a-settings-icon-to-the-left-of-the-sign-in-sign-out-icon-c-BT2DRbCjzO-jWajzdwjbyjC

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,11 @@
 	"contributes": {
 		"commands": [
 			{
+				"command": "intuita.globalConfigSettings",
+				"title": "Configure global settings",
+				"icon": "$(settings-gear)"
+			},
+			{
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,6 +38,11 @@
 	},
 	"contributes": {
 		"commands": [
+			{
+				"command": "intuita.globalConfigSettings",
```
grzpab, line 42: can you rename it to `showIntuitaSettings`, these are not necessarily global settings (user vs workspace)

hbjORbj: addressed

PR #823, opened 2023-09-20
INT-1899 Welcome message for private registry

hbjORbj:
Screenshot 2023-09-20 at 12 05 36


grzpab (commented):
Can you ensure you use the same font (size included) as it is used in e.g. Codemod Runs empty labels?

hbjORbj:
Font style are the same. I changed the font-size to match other welcome messages.

Screenshot 2023-09-20 at 12 17 17

PR #825, opened 2023-09-20
INT-1898 Use custom notifs for auth flow

hbjORbj:
claap: https://app.claap.io/intuita/use-custom-notifs-for-auth-flow-c-BT2DRbCjzO-Ud4kv8oT7yLo

### DIFF
--- a/intuita-webview/src/main/App.tsx
+++ b/intuita-webview/src/main/App.tsx
@@ -1,6 +1,8 @@
+import { toast } from 'react-toastify';
 import { useEffect, useRef, useState } from 'react';
 
 import {
+	VSCodeButton,
 	VSCodePanels,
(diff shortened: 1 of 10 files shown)


intuita-webview/src/main/App.tsx
```diff
@@ -18,6 +20,17 @@ import { ToastContainer } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
 import { useTheme } from '../shared/Snippet/useTheme';
 
+const ToastContainerProps = {
```
DmytroHryshyn, line 23: can you rename it to `toastContainerProps` (lowercase)

hbjORbj: addressed, thanks

PR #832, opened 2023-11-30
feat: INT-2193 delete cases directory

r4zendev:


package.json
```diff
@@ -250,7 +250,7 @@
 		"install:webview": "cd intuita-webview && pnpm install",
 		"build:webview": "cd intuita-webview && pnpm build",
 		"watch:webview": "cd intuita-webview && pnpm start",
-		"package": "vsce package --no-dependencies"
+		"package": "vsce package --no-dependencies --no-yarn"
```
grzpab, line 253: let's see if it works

src/components/fileService.ts
```diff
@@ -82,6 +92,17 @@ export class FileService {
 		await this.deleteFiles({ uris: [params.oldUri] });
 	}
 
+	public async deleteDirectories(params: {
+		uris: ReadonlyArray<Uri>;
+	}): Promise<void> {
+		for (const uri of params.uris) {
+			await workspace.fs.delete(uri, {
```
grzpab, line 99: can you wrap it with try/catch, it might fail when the directory doesn't exist.

grzpab (approved):

PR #843, opened 2023-12-16
chore: support all job kinds from the streamed cases

grzpab:


grzpab:
@alex-from-intuita as discussed, I will merge and release it.

PR #850, opened 2024-01-03
INT-2373 do not initialize the VSCE if there is no active workspace

grzpab:
image

### DIFF
--- a/intuita-webview/src/main/App.tsx
+++ b/intuita-webview/src/main/App.tsx
@@ -90,7 +90,10 @@ function App() {
 	}, []);
 
 	useEffect(() => {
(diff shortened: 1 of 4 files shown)


hbjORbj (approved):

intuita-webview/src/main/App.tsx
```diff
@@ -131,6 +134,16 @@ function App() {
 		});
 	};
 
+	if (mainWebviewViewProps === null) {
+		return (
+			<main className="App" ref={ref}>
+				<p className="warning">
+					Open a workspace folder to use the Intuita VSCode Extension.
```
hbjORbj, line 141: maybe remove `folder`?

PR #851, opened 2024-01-03
INT-2372 Fix the selection color for light theme

hbjORbj:
Screenshot 2024-01-03 at 11 08 53
Screenshot 2024-01-03 at 11 09 14

### DIFF
--- a/intuita-webview/src/codemodList/CodemodNodeRenderer/style.module.css
+++ b/intuita-webview/src/codemodList/CodemodNodeRenderer/style.module.css
@@ -44,6 +44,10 @@
 	align-items: center;
 }
 
+.labelContainer.focused {
+	color: var(--vscode-list-activeSelectionForeground);
+}
+
 .directorySelector {
 	display: flex;
 }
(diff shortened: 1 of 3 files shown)


intuita-webview/src/codemodList/CodemodNodeRenderer/style.module.css
```diff
@@ -44,6 +44,10 @@
 	align-items: center;
 }
 
+.labelContainer.focused {
```
grzpab, line 47: can we use the https://developer.mozilla.org/en-US/docs/Web/CSS/:focus pseudo-class here?

DmytroHryshyn (approved):

intuita-webview/src/codemodList/CodemodNodeRenderer/style.module.css
hbjORbj: Nope. We are not handling pseudo classes well in the app. Good point because this is something we use improve for better a11y

grzpab (approved):

PR #859, opened 2024-01-25
Add Windows support to VSCE

DmytroHryshyn:
https://app.claap.io/intuita/c-BT2DRbCjzO-8IF91qEU9YBz

### DIFF
--- a/src/components/bootstrapExecutablesService.ts
+++ b/src/components/bootstrapExecutablesService.ts
@@ -70,8 +70,8 @@ export class BootstrapExecutablesService {
 			process.platform === 'darwin'
 				? 'macos'
 				: encodeURIComponent(process.platform);
-		
-		if(platform === 'win32') {
+
+		if (platform === 'win32') {
 			return null;
 		}
 
--- a/src/components/bootstrapExecutablesService.ts
+++ b/src/components/bootstrapExecutablesService.ts
@@ -53,13 +53,13 @@ export class BootstrapExecutablesService {
 				'755',
 			);
 		} catch (error) {
-			if (!(error instanceof ForbiddenRequestError)) {
-				throw error;
-			}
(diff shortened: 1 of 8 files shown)


src/components/bootstrapExecutablesService.ts
```diff
@@ -65,12 +65,16 @@ export class BootstrapExecutablesService {
...
 
-	private async __bootstrapCodemodEngineRustExecutableUri(): Promise<Uri> {
+	private async __bootstrapCodemodEngineRustExecutableUri(): Promise<Uri | null> {
 		const platform =
 			process.platform === 'darwin'
 				? 'macos'
 				: encodeURIComponent(process.platform);
 
+		if (platform === 'win32') {
+			return null;
```
grzpab, line 75: I think we could compile the Rust app for Windows as well, I just couldn't cross compile it with Linux.

src/components/engineService.ts
```diff
@@ -275,16 +291,8 @@ export class EngineService {
...
-
-		if (executableUri === null) {
-			throw new EngineNotFoundError(
-				'The codemod engine node has not been downloaded yet',
-			);
-		}
-
 		const childProcess = spawn(
-			singleQuotify(executableUri.fsPath),
+			this.__getCodemodEngineNodeExecutableCommand(),
```
grzpab, line 295: will it always deliver a correct path?

DmytroHryshyn: it can throw if engineNode is not bootstrapped, in other case it should return exactly same path as previously

src/components/bootstrapExecutablesService.ts
DmytroHryshyn: but not in current release, right?

grzpab: no, not yet :)

grzpab (approved):

PR #863, opened 2024-01-31
Feat/add axios retry and better errors

DmytroHryshyn:


src/axios/index.ts
```diff
@@ -0,0 +1,8 @@
+import axios from 'axios';
+import axiosRetry from 'axios-retry';
+
+const client = axios.create();
+axiosRetry(client);
+
+export const DEFAULT_RETRY_COUNT = 3;
+export default client;
```
grzpab, line 8: can we call it `retryingClient` and export it non-defaultly?

package.json
grzpab, line 1: License check (no changes, the "red" libraries have been already discussed)
image

test/dowloadService.test.ts
```diff
@@ -0,0 +1,83 @@
...
+		await downloadService.downloadFileIfNeeded(
+			`https://test.com/test`,
+			// @ts-expect-error passing a string instead of URI, because URI cannot be imported from vscode
+			'/',
+			'755',
+		);
+
+		expect(fs.writeFile).toBeCalledWith(
+			'/',
+			new Uint8Array([84, 101, 115, 116]),
```
grzpab, line 80: what does that mean?

DmytroHryshyn: We mock `get` response with `Test`, and expect that writeFile will be called with uint representation of that string. fs.writeFile is mocked, so we can check args it was called with

grzpab: Ah ok, the array in question is basically `Test`:
```
new Uint8Array([84, 101, 115, 116]).forEach((x) => console.log(String.fromCharCode(x)))
```

grzpab (approved):

